PR #2, opened 2021-04-28
Pass urlFactory to SuggestionsLoading

mallexxx:
Task/Issue URL: https://app.asana.com/0/inbox/1199237043628108/1200180252402997/1200259155963726
Tech Design URL:
CC: @tomasstrba 

**Description**:
This PR adds ability to parse Suggestion Phrases into .website(url:) providing an extra urlFactory: (String) -> URL? callback

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14

* [ ] macOS 10.15
* [x] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Suggestions/SuggestionLoading.swift
+++ b/Sources/BrowserServicesKit/Suggestions/SuggestionLoading.swift
@@ -22,18 +22,9 @@ public protocol SuggestionLoading: AnyObject {
 
     func getSuggestions(query: Query,
                         maximum: Int,
-                        urlFactory: @escaping (String) -> URL?,
                         completion: @escaping ([Suggestion]?, Error?) -> Void)
 
     var dataSource: SuggestionLoadingDataSource? { get set }
-
-}
-
-extension SuggestionLoading {
-
-    func getSuggestions(query: Query, maximum: Int, completion: @escaping ([Suggestion]?, Error?) -> Void) {
-        getSuggestions(query: query, maximum: maximum, urlFactory: { _ in nil }, completion: completion)
-    }
 
 }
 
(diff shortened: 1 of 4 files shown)


tomasstrba (commented):
What do you think of passing urlFactory as a parameter of SuggestionLoader init method instead of passing the same block every time we fetch suggestions?

mallexxx:
@tomasstrba yeah, makes sense!

Sources/BrowserServicesKit/Suggestions/SuggestionLoading.swift
```diff
@@ -38,9 +40,11 @@ public class SuggestionLoader: SuggestionLoading {
     }
 
     public weak var dataSource: SuggestionLoadingDataSource?
+    private var urlFactory: ((String) -> URL?)?
 
-    public init(dataSource: SuggestionLoadingDataSource? = nil) {
+    public init(dataSource: SuggestionLoadingDataSource? = nil, urlFactory: ((String) -> URL?)? = nil) {
```
tomasstrba, line 45: Sorry, just few details that got my attention. What is the reason to have the urlFactory as optional? I can imagine it can be overlooked with the nil as default argument and the library will basically not work correctly.

mallexxx: This is done to keep the API backwards compatible

brindy: It's pretty early in the life of this library and the usage of it, let's not worry about that so much.  Breaking API changes just mean bumping the major version so if we rapidly move from 1.0.1 to 2.0.0 then that's fine, that's the semver.org way and nothing to worry about. :)

mallexxx:
Updated to pass urlFactory in init as non-optional and use self.urlFactory in remoteSuggestions

tomasstrba (approved):
LGTM! 👍

PR #10, opened 2021-06-10
Add autofill prerelease changes

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1200425975637283/f
Tech Design URL:
CC:

**Description**:

This PR adds changes for the prerelease autofill feature.

**Steps to test this PR**:
1. See Asana task

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


Sources/BrowserServicesKit/Email/EmailKeychainManager.swift
```diff
@@ -51,6 +51,46 @@ extension EmailKeychainManager: EmailManagerStorage {
...
+    public func deleteInviteCode() {
+        EmailKeychainManager.deleteItem(forField: .inviteCode)
+    }
+
+    public func deleteWaitlistState() {
+        EmailKeychainManager.deleteWaitlistState()
+    }
 }
 
 private extension EmailKeychainManager {
```
bwaresiak, line 96: General comment on Keychain storage: due to nature of the Keychain (storing outside of app sandbox) entries are not removed upon app reinstall. This differs from Android implementation and leaves potential concerns about "clean state" of the new installation - this most likely should be addressed, probably either by: 
 - Moving from keychain to other secure storage within app container.
 - Pairing key with UUID generated for every new installation so we don't reuse old ones (but that would leave orphaned entries there and cleanup is needed in such case).
 - Cleaning up keychain when app is being (re)installed (we'd need to set a flag in UserDefaults for that - and hope that there is no access issues when cold-starting that would wipe out the keychain by accident)

Sources/BrowserServicesKit/Email/EmailKeychainManager.swift
```diff
@@ -127,6 +186,12 @@ private extension EmailKeychainManager {
         deleteItem(forField: .token)
         deleteItem(forField: .alias)
     }
+
+    static func deleteWaitlistState() {
```
bwaresiak, line 190: 1. Shouldn't this be part of `deleteAll()` ?
2. I see it only used from debug screen - in case we have implementation that is "dead" on production it would be good to have clear separation of concerns, at least through extensions.

samsymons: `deleteAll` should be renamed, it's used to log out whereas the waitlist state line is only used by the debug VC as you mentioned. I'll try make this more clear. :+1:

samsymons: Renamed `deleteAll` to `deleteAuthenticationState` for clarity, and have moved the functions that are only used for debugging into a demarcated extension at the bottom of the file.

bwaresiak (approved):

Sources/BrowserServicesKit/Email/EmailKeychainManager.swift
```diff
@@ -98,31 +141,49 @@ private extension EmailKeychainManager {
...
+        deleteItem(forField: field)
+        add(data: stringData, forField: field)
     }
     
     static func add(data: Data, forField field: EmailKeychainField) {
         let query: [String: Any] = [
             kSecClass as String: kSecClassGenericPassword,
             kSecAttrSynchronizable as String: false,
             kSecAttrService as String: field.keyValue,
+            kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlock,
```
samsymons, line 180: Added this attribute to facilitate background refresh.

PR #11, opened 2021-06-15
don't inject autofill into frames on pre-ios 14

brindy:
Task/Issue URL: https://app.asana.com/0/inbox/392891325557408/1200465696711800/1200466433503758
Tech Design URL:
CC: @GioSensation 

**Description**:

Messaging into frames isn't supported before iOS 11, macOS 14 and causes unexpected side effects when trying to inject autofill there.

**Steps to test this PR**:
1. On iOS 13 - using 4.0.1 of BSK visit the SERP with a simulator and debug console active.  You'll see a console error and jext pixel.
1. Update to using this version of BSK, the console and jext pixel will be gone.
1. Ensure autofill still works as expected on both iOS 13 and 14

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Autofill/AutofillUserScript.swift
+++ b/Sources/BrowserServicesKit/Autofill/AutofillUserScript.swift
@@ -62,7 +62,13 @@ public class AutofillUserScript: NSObject, UserScript {
     }()
 
     public var injectionTime: WKUserScriptInjectionTime { .atDocumentStart }
-    public var forMainFrameOnly: Bool { false }
+    public var forMainFrameOnly: Bool {
+        if #available(iOS 14, macOS 11, *) {
+            return false
+        }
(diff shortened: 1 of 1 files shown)


Sources/BrowserServicesKit/Autofill/AutofillUserScript.swift
```diff
@@ -191,11 +197,11 @@ extension AutofillUserScript {
 
             let script = """
             (() => {
-                window.\(methodName)({
+                window.\(methodName) && window.\(methodName)({
```
brindy, line 200: Ensure we only call the secret method if it's actually available.  This is more of a fall back position just in case, to avoid creating errors in the console.

GioSensation (approved):
Looks good. Thanks! :shipit:

PR #34, opened 2021-11-22
Update the credit card table

samsymons:
Task/Issue URL: https://app.asana.com/0/1199230911884351/1201406179271672/f
Tech Design URL:
CC: @brindy 

**Description**:

This PR updates the credit card table to store the number as L2 data.

There are a couple other small database tweaks included here too, such as storing an additional address column.

**Steps to test this PR**:

⚠️ Back up your existing Vault.db file, in case something in this migration is broken.
Go to `~/Library/Containers/com.duckduckgo.macos.browser.debug/Data/Library/Application Support/Vault` to find the Vault file, and copy/paste it somewhere else for the time being.

**This PR requires checking out the corresponding desktop branch.**

1. Run this branch with existing Vault data, and check that the migration completes successfully. Ideally, run this on existing Vault data that already has some credit cards
1. Create a credit card and check that it shows up correctly when Autofilling; you can easily test this somewhere like Amazon.com
1. Edit an existing card and check that it has updated correctly

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/SecureVaultFactory.swift
+++ b/Sources/BrowserServicesKit/SecureVault/SecureVaultFactory.swift
@@ -67,11 +67,10 @@ public class SecureVaultFactory {
     }
     
     internal func makeSecureVaultProviders() throws -> SecureVaultProviders {
-        let cryptoProvider = makeCryptoProvider()
-        let keystoreProvider = makeKeyStoreProvider()
-        let databaseProvider: SecureVaultDatabaseProvider
-        
         do {
+            let (cryptoProvider, keystoreProvider) = try createAndInitializeEncryptionProviders()
+            let databaseProvider: SecureVaultDatabaseProvider
+
             if let existingL1Key = try keystoreProvider.l1Key() {
                 databaseProvider = try DefaultDatabaseProvider(key: existingL1Key)
             } else {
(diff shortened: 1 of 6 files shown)


Sources/BrowserServicesKit/SecureVault/SecureVaultFactory.swift
```diff
@@ -96,5 +104,26 @@ public class SecureVaultFactory {
...
+            let passwordKey = try cryptoProvider.deriveKeyFromPassword(password)
+            let encryptedL2Key = try cryptoProvider.encrypt(l2Key, withKey: passwordKey)
+
+            try keystoreProvider.storeEncryptedL2Key(encryptedL2Key)
+            try keystoreProvider.storeGeneratedPassword(password)
+            try keystoreProvider.storeL1Key(l1Key)
+            
+            return (cryptoProvider, keystoreProvider)
+        }
+    }
```
samsymons, line 127: Not super satisfied with this atm. Database migrations only need the crypto and key store providers, and creating the database provider anyway causes a failed migration as it tries to migrate... while already migrating.

Open for feedback on a cleaner way to handle this.

brindy: Yeah, let's chat about this as that's a big chunk of duplicated code that I think only needs to exist here.

brindy (approved):
LGTM subject to the wee bug we just chatted about.

PR #43, opened 2021-12-24
Add JS build step to increase code reuse

jonathanKingston:
Task/Issue URL:
Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


jonathanKingston:
I think this has bitrotted enough we should drop it.
I think in general however we should shift as much JS as possible into content-scope-scripts which has this tooling already.
I'm happy to advise bringing back this kinda change if we want to retain JS reuse in this codebase.

PR #44, opened 2021-12-27
Support content scope as a user script

jonathanKingston:
Task/Issue URL: https://app.asana.com/0/1163321984198618/1201071474741984/f
Tech Design URL:
CC:

**Description**:

Loads the code that is currently in the extension as a user script that's bundled and accounts for the privacy config.

https://github.com/duckduckgo/content-scope-scripts is the bundled code.

This change doesn't implement anything in isolation so would be safe to merge, the scripts are controlled by remote config and require integration into iOS etc.

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/UserScript/ContentScopeUserScript.swift
+++ b/Sources/BrowserServicesKit/UserScript/ContentScopeUserScript.swift
@@ -20,23 +20,38 @@ import Foundation
 import WebKit
 import Combine
 
+public final class ContentScopePreferences: Encodable {
+    public let globalPrivacyControlValue: Bool
+    public let debug: Bool = false
+    public let sessionKey: String = "TODO"
+
+    public init(gpcEnabled: Bool) {
+        self.globalPrivacyControlValue = gpcEnabled
+    }
+}
+
 public final class ContentScopeUserScript: NSObject, UserScript {
     public let messageNames: [String] = []
 
-    public init(_ privacyConfigManager: PrivacyConfigurationManager) {
-        source = ContentScopeUserScript.generateSource(privacyConfigManager)
+    public init(_ privacyConfigManager: PrivacyConfigurationManager, preferences: ContentScopePreferences) {
+        source = ContentScopeUserScript.generateSource(privacyConfigManager, preferences: preferences)
     }
 
-    public static func generateSource(_ privacyConfigurationManager: PrivacyConfigurationManager) -> String {
+    public static func generateSource(_ privacyConfigurationManager: PrivacyConfigurationManager, preferences: ContentScopePreferences) -> String {
         let privacyConfigJson = String(data: privacyConfigurationManager.currentConfig, encoding: .utf8)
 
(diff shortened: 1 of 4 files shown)


Sources/BrowserServicesKit/UserScript/ContentScopeUserScript.swift
```diff
@@ -0,0 +1,68 @@
...
+              let jsonPreferences = try? JSONEncoder().encode(preferences),
+              let jsonPreferencesString = String(data: jsonPreferences, encoding: .utf8)
+              else {
+            return ""
+        }
+        
+        return loadJS("contentScope", from: Bundle.module, withReplacements: [
+            "$CONTENT_SCOPE$": privacyConfigJson!,
+            "$USER_UNPROTECTED_DOMAINS$": jsonString,
+            "$USER_PREFERENCES$": jsonPreferencesString,
```
shakyShane, line 55: @jonathanKingston one thing I'm not 100% clear on here, is how each platform will pass platform-specific configuration to the content scripts?

For example, if we want the future navigator interface script to have access to `{ platform: { name: "ios" }` on iOS and `{ platform: { name: "macos" }`, where would this design allow that?

jonathanKingston: Either:

- If we need it remotely configurable through `privacyConfigJson` which is the remote config raw JSON file and we'd parse that out in the relevant feature file.
- Or we can pass it through `ContentScopePreferences` for constants like you're referring to.

The platform itself is responsible for the relevant `inject/platform.js` script and that will likely be different in how this data gets wired into the calling API of `contentScopeFeatures.init()` (eventually this will be typescript checked so it's clear what this API is)

jonathanKingston: https://github.com/duckduckgo/content-scope-scripts/blob/main/inject/apple.js#L53 is where these call sites go in and this combines the data in the format expected by `contentScopeFeatures.init(processedConfig)`

Sources/BrowserServicesKit/UserScript/ContentScopeUserScript.swift
```diff
@@ -0,0 +1,68 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import WebKit
+import Combine
+
+public final class ContentScopePreferences: Encodable {
```
jonathanKingston, line 23: Perhaps "properties" is a less confusing name, it's essentially a transport for platform constants and user preferences?

jonathanKingston:
Closing this in favour of: https://github.com/duckduckgo/BrowserServicesKit/pull/45 as it doesn't rely on the build script PR.

PR #49, opened 2022-01-11
Remove content scope files causing warnings

jonathanKingston:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1201640066177016/f
Tech Design URL:
CC:

**Description**:

- This removes the warnings when building.
- I couldn't find a way to make this work without requiring all the hard coded files (without turning the content scope scripts into a package which we probably will want long term https://app.asana.com/0/1201614831475344/1201638298093318/f)

**Steps to test this PR**:
1. Open up XCode and build
2. Check for no warnings

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


samsymons:
>I couldn't find a way to make this work without requiring all the hard coded files

No worries, we've looked at that before and couldn't find a way either, hence the list of Autofill files in there too 🙂

Package.swift
```diff
@@ -25,6 +25,19 @@ let package = Package(
                 "TrackerRadarKit"
             ],
             exclude: [
+                "Resources/content-scope-scripts/README.msd",
```
samsymons, line 28: Small typo here:

```suggestion
                "Resources/content-scope-scripts/README.md",
```

samsymons (changes requested):
One thing to fix here and then this is good to go, all warnings are resolved after this :clap:

It'd be ideal if Bitrise would fail on legitimate warnings so that we can catch them easier in the future, but I don't think that's in scope for this PR.

Thanks for fixing this!

samsymons (approved):
LGTM!

PR #53, opened 2022-01-20
Multiple blocking rule lists

SlayterDev:
Task/Issue URL: https://app.asana.com/0/0/1201694797383792/f
Tech Design URL:
CC: @bwaresiak @brindy @ladamski 

**Description**:
This PR adds support for multiple content blocking rules

**Steps to test this PR**:
1. See testing in FB click to load PR
1.

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/ContentBlocking/ContentBlockerRulesManager.swift
+++ b/Sources/BrowserServicesKit/ContentBlocking/ContentBlockerRulesManager.swift
@@ -25,9 +25,107 @@ import TrackerRadarKit
 public protocol ContentBlockerRulesUpdating {
 
     func rulesManager(_ manager: ContentBlockerRulesManager,
-                      didUpdateRules: ContentBlockerRulesManager.CurrentRules,
+                      didUpdateRules: [ContentBlockerRulesManager.Rules],
                       changes: ContentBlockerRulesIdentifier.Difference,
                       completionTokens: [ContentBlockerRulesManager.CompletionToken])
+}
+
+/**
+ Encapsulates compilation steps for a single Task
+ */
+private class CompilationTask {
+    typealias Completion = () -> Void
+    let workQueue: DispatchQueue
+    let rulesList: ContentBlockerRulesList
+    let sourceManager: ContentBlockerRulesSourceManager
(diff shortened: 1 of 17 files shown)


Sources/BrowserServicesKit/ContentBlocking/ContentBlockerRulesManager.swift
```diff
@@ -25,11 +25,109 @@ import TrackerRadarKit
...
+    }
+
+    private func compilationFailed(for model: ContentBlockerRulesSourceModel,
+                                   with error: Error,
+                                   completionHandler: @escaping Completion) {
+        os_log("Failed to compile %{public}s rules %{public}s", log: logger, type: .error, rulesList.name, error.localizedDescription)
+
+        // Retry after marking failed state in the source
+        sourceManager.compilationFailed(for: model, with: error)
+        let newModel = sourceManager.makeModel()
```
mallexxx, line 91: I think this should go inside self.workQueue.async { .. } as sourceManager.makeModel() usage is made from the workQueue and the compilationFailed callback may be called by the compileContentRuleList from the main thread

mallexxx (approved):
Looks good to me (see minor data race comment inline)
CompilationTask concept still looks over-complicated, but this to be targeted by my next ContentRules PR

Sources/BrowserServicesKit/ContentBlocking/ContentBlockerRulesManager.swift
bwaresiak: It's not like there's a bug - as operations here are done sequentially and are encapsulated to a single class instance - but this is something I *know* as an author of this class :) and it isn't necessary a good thing. I agree that having this in a async block is more clear to the reader and possibly less error prone in case of any refactoring - going to change this.

ladamski:
Are we good to merge?  Looks like it but want to check in before I pull the trigger :)

bwaresiak:
@ladamski I believe so :)

PR #57, opened 2022-02-02
Enable unit tests via Xcode

samsymons:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1201743270866809/f
Tech Design URL:
CC: @tomasstrba 

**Description**:

This PR updates BSK such that unit tests can run both via `swift test` and from within Xcode.

I'm not clear on why this differs between Xcode and `swift test`, but using the `.copy` rule preserves the directory structure of our resources and requires that path to be specified in the bundle. For example, leaving `.copy` in place means that we would need to prepend `Resources/` to the paths when looking them up during test runs.

Rather than do this, I've changed the rules to `.process` which turns the bundle into a flat structure.

I also removed a redundant test file, which we had two copies of under the same name but in two different directories.

**Steps to test this PR**:
1. Check out this branch
2. Run tests via `swift test`
3. Run tests via Xcode

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


Package.swift
```diff
@@ -65,9 +65,8 @@ let package = Package(
                 "BrowserServicesKit"
             ],
             resources: [
-                .copy("UserScript/testUserScript.js"),
-                .copy("MockFiles"),
```
samsymons, line 69: `MockFiles` only contains `privacy-config-example.json` which is duplicated [here](https://github.com/duckduckgo/BrowserServicesKit/blob/03b42987d765e92b58a5975f6f6edad503100ec8/Tests/BrowserServicesKitTests/Resources/privacy-config-example.json). I compared these two files with a diff tool, they're the same so let's remove one. 🙂

mallexxx (approved):
LGTM!

PR #60, opened 2022-02-03
Recreate SecureVault database on corruption

mallexxx:
Task/Issue URL: https://app.asana.com/0/1177771139624306/1201649423461236/f
Tech Design URL:
CC: @samsymons 

**Description**:
Tries to recreate database file on corruption

see https://github.com/more-duckduckgo-org/macos-browser/pull/417

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/SecureVaultDatabaseProvider.swift
+++ b/Sources/BrowserServicesKit/SecureVault/SecureVaultDatabaseProvider.swift
@@ -61,20 +61,17 @@ final class DefaultDatabaseProvider: SecureVaultDatabaseProvider {
 
     let db: DatabaseQueue
 
-    init(key: Data) throws {
+    init(file: URL = DefaultDatabaseProvider.dbFile(), key: Data) throws {
         var config = Configuration()
         config.prepareDatabase {
             try $0.usePassphrase(key)
         }
 
-        let file = try Self.dbFile()
-
         do {
             db = try DatabaseQueue(path: file.path, configuration: config)
         } catch (let error) {
-            os_log("database initialization failed with %{public}s, recreating", type: .error, error.localizedDescription)
-            try? FileManager.default.removeItem(at: file)
-            db = try DatabaseQueue(path: file.path, configuration: config)
+            os_log("database initialization failed with %{public}s", type: .error, error.localizedDescription)
+            throw error
         }
 
         var migrator = DatabaseMigrator()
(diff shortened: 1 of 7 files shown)


Sources/BrowserServicesKit/SecureVault/SecureVaultDatabaseProvider.swift
```diff
@@ -67,7 +68,14 @@ final class DefaultDatabaseProvider: SecureVaultDatabaseProvider {
         }
 
         let file = try Self.dbFile()
-        db = try DatabaseQueue(path: file.path, configuration: config)
+
+        do {
+            db = try DatabaseQueue(path: file.path, configuration: config)
```
bwaresiak, line 73: Is it possible for DB initialization to fail due to lack of storage space (for temp files)? If that is the case, then removing DB would basically be a bug and data loss for the client. Question is whether there are errors that we can recover from or that are temporary - in such case removing DB may not be the best course of action.

mallexxx: Good point, I've added extra steps to validate a new database can be created without dropping the old one: it's renamed to Vault[_N].db.bak after we make sure we can create new database files which should validate we're not dealing with write permissions or write errors and keep users data backup just in case

bwaresiak: I'd assume there is a finite set of errors we can have when initializing DB - some may be obvious, some may be weird, but generally we should be able to split that set into two paths:
 - Non-recoverable failure - we need to create new DB file as you do here.
 - Recoverable failure - this is product decision what to do with this, we may want to show some dialog to the user (e.g. out of space info).

I don't think we need to solve the latter here, but we should track DB initialization errors (as we do on iOS) and fire error pixel along with error code (and possibly underlying error if available) to understand frequency of this issue and segment it somehow. 

So I suggest:
 - Add error pixels.
 - Create Grafana board for that (I can do this once pixel is there)
 - Schedule a follow up task (2 weeks? maybe before release date?) to validate error types and see if we can improve this solution and make sure we are not breaking something by accident. 

What do you think?

mallexxx:
Changed recreation to happen only on db corruption and implemented error reporting

samsymons (approved):
LGTM too :+1:

PR #75, opened 2022-03-04
Ensure that URL.addParameter only encodes the parameter being added

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/1177771139624306/1201865561825271/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1082
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/456

**Description**:
`URL.addParameter` used to call `encodingPluses` on the whole URL query, which had a side effect of potentially changing existing query parameters. I updated the function so that it:
* temporarily replaces existing query with only the parameter being added,
* encodes that parameter alone,
* merges existing encoded query with the newly added encoded parameter.

**Steps to test this PR**:
Note that the bug is only present on MacOS.
1. Search for `Battlestar Galactica`
1. Click in the address field
1. Hit enter
1. The search should be resubmitted, instead of changing to `Battlestar+Galactica`

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
+++ b/Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
@@ -74,12 +74,17 @@ extension URL {
 
     public func addParameter(name: String, value: String) throws -> URL {
         guard var components = URLComponents(url: self, resolvingAgainstBaseURL: false) else { throw ParameterError.parsingFailed }
-        var queryItems = components.queryItems ?? [URLQueryItem]()
+        var encodedQueryItems = components.percentEncodedQueryItems ?? [URLQueryItem]()
+
         let newQueryItem = URLQueryItem(name: name, value: value)
-        queryItems.append(newQueryItem)
-        components.queryItems = queryItems
+        components.queryItems = [newQueryItem]
         guard let encodedQuery = components.percentEncodedQuery else { throw ParameterError.encodingFailed }
         components.percentEncodedQuery = encodedQuery.encodingPluses()
+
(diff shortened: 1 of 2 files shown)


Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
```diff
@@ -74,12 +74,17 @@ extension URL {
 
     public func addParameter(name: String, value: String) throws -> URL {
         guard var components = URLComponents(url: self, resolvingAgainstBaseURL: false) else { throw ParameterError.parsingFailed }
-        var queryItems = components.queryItems ?? [URLQueryItem]()
+        var encodedQueryItems = components.percentEncodedQueryItems ?? [URLQueryItem]()
+
         let newQueryItem = URLQueryItem(name: name, value: value)
-        queryItems.append(newQueryItem)
-        components.queryItems = queryItems
+        components.queryItems = [newQueryItem]
```
ayoy, line 80: Instead of doing this, we could probably just call `addingPercentEncoding()` on the `name` and `value`. This would make the implementation simpler, but in my approach we rely on `URLComponents` to perform percent encoding. I belive it's a bit safer approach (more future-proof since we expect `URLComponents` to do the encoding properly), but please let me know if I'm being too cautious.

brindy (approved):
LGTM! Very nice.  After merging to main, since there were no API changes, please increment the version by 0.0.1.

PR #76, opened 2022-03-10
Prompt to store cards and identities

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1201938691568505/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1086
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/461

**Optional**:

CC: @brindy 

**Description**:

This PR updates BrowserServicesKit to handle the new `pmHandlerStoreData` call. As a part of this, it runs some logic to look up whether the provided payment method or address already exist in the database.

**Steps to test this PR**:
1. See macOS PR for more information

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/SecureVault.swift
+++ b/Sources/BrowserServicesKit/SecureVault/SecureVault.swift
@@ -46,12 +46,14 @@ public protocol SecureVault {
 
     func identities() throws -> [SecureVaultModels.Identity]
     func identityFor(id: Int64) throws -> SecureVaultModels.Identity?
+    func existingIdentityForAutofill(matching proposedIdentity: SecureVaultModels.Identity) throws -> SecureVaultModels.Identity?
     @discardableResult
     func storeIdentity(_ identity: SecureVaultModels.Identity) throws -> Int64
     func deleteIdentityFor(identityId: Int64) throws
 
     func creditCards() throws -> [SecureVaultModels.CreditCard]
     func creditCardFor(id: Int64) throws -> SecureVaultModels.CreditCard?
+    func existingCardForAutofill(matching proposedCard: SecureVaultModels.CreditCard) throws -> SecureVaultModels.CreditCard?
     @discardableResult
     func storeCreditCard(_ card: SecureVaultModels.CreditCard) throws -> Int64
     func deleteCreditCardFor(cardId: Int64) throws
@@ -275,6 +277,14 @@ class DefaultSecureVault: SecureVault {
             try self.providers.database.deleteIdentityForIdentityId(identityId)
         }
     }
+    
+    func existingIdentityForAutofill(matching proposedIdentity: SecureVaultModels.Identity) throws -> SecureVaultModels.Identity? {
+        let identities = try self.identities()
+        
+        return identities.first { existingIdentity in
+            existingIdentity.hasAutofillEquality(comparedTo: proposedIdentity)
+        }
+    }
 
     // MARK: - Credit Cards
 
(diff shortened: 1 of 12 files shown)


Sources/BrowserServicesKit/SecureVault/SecureVaultManager.swift
```diff
@@ -158,5 +150,64 @@ extension SecureVaultManager: AutofillSecureVaultDelegate {
...
+                }
+
+            } else {
+                os_log("Received new credentials to save", log: .passwordManager, type: .info)
+                let account = SecureVaultModels.WebsiteAccount(username: credentials.username ?? "", domain: domain)
+                proposedCredentials = SecureVaultModels.WebsiteCredentials(account: account, password: passwordData)
+            }
+        } else {
+            os_log("No new credentials found, avoid prompting user", log: .passwordManager, type: .info)
+        }
```
samsymons, line 195: This has gotten a bit out of hand, I'll try to clean it up. The Autofill functionality has changed in such a way that it doesn't try to protect against storeData calls directly after you've autofilled, which means BSK now has to do duplicate checks around credentials.

Sources/BrowserServicesKit/SecureVault/SecureVault.swift
```diff
@@ -275,6 +277,14 @@ class DefaultSecureVault: SecureVault {
             try self.providers.database.deleteIdentityForIdentityId(identityId)
         }
     }
+    
+    func existingIdentityForAutofill(matching proposedIdentity: SecureVaultModels.Identity) throws -> SecureVaultModels.Identity? {
+        let identities = try self.identities()
+        
+        return identities.first { existingIdentity in
+            existingIdentity.hasAutofillEquality(comparedTo: proposedIdentity)
```
samsymons, line 285: It would be preferable to store the normalized Autofill identifiers in the database and look them up directly rather than having to pull all of the objects and check them each in linear time, but the count of this array is typically very small (i.e., < 10) so the simpler approach is used here. Can be optimized for performance if we need it later.

Sources/BrowserServicesKit/SecureVault/SecureVaultModels.swift
```diff
@@ -197,20 +226,72 @@ public struct SecureVaultModels {
...
+        public var formattedName: String {
+            return Self.mediumPersonNameComponentsFormatter.string(from: nameComponents)
+        }
+        
+        public var longFormattedName: String {
+            return Self.longPersonNameComponentsFormatter.string(from: nameComponents)
+        }
+        
+        var autofillEqualityName: String?
+        var autofillEqualityAddressStreet: String?
```
samsymons, line 260: Need to improve this, it's not great since initializers in other files are having to ensure that they set these values too, making it hard to write a new initializer and know that you're doing it correctly.

ayoy: Right, making these vars lazy or turning them into functions backed by a private var that is initialized on demand makes things even more complicated, because `SecureVaultAutofillEquatable` needs to define its function `mutable` and taking `inout` parameter.

There is probably no nice solution for handling it behind the scenes.

ayoy (approved):
LGTM, solid effort! I added a single-line change that fixes deserializing `IncomingCredentials` from JSON.

Sources/BrowserServicesKit/SecureVault/SecureVaultManager.swift
```diff
@@ -158,5 +150,64 @@ extension SecureVaultManager: AutofillSecureVaultDelegate {
...
         }
     }
+    
+    func existingEntries(for domain: String, autofillData: AutofillUserScript.DetectedAutofillData) throws -> AutofillData {
+        
+        let vault = try SecureVaultFactory.default.makeVault(errorReporter: self.delegate)
+        
+        // Determine if the identity should be sent to the client app:
+
+        var proposedIdentity: SecureVaultModels.Identity?
```
ayoy, line 160: I know this isn't groundbreaking, but to increase readability you could extract the code responsible for retrieving proposed identity, credentials and credit card into 3 helper methods. It would make this function less scary at first sight :) I believe it's also the right thing to have in case we'll have more autofill types going forward.

PR #80, opened 2022-04-01
Conditionally allow reserved characters in URL.addParameter

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/1163321984198618/1202033667108435/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1108
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/498

**Description**:
Add an optional argument to `URL.addParameter` to specify reserved characters (as per [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.2) that should not be percent-encoded, for example because they play a special role in the payload, as defined by the API provider.

By default we percent-encode all reserved characters, which should be a safe default, unless specific API providers instruct to allow certain characters. This is the case for Website Breakage reports where comma is used to delimit blocked trackers domains.

**Steps to test this PR**:
This is an macOS-only change, as iOS does not use the `URL.addParameter` at this time. Furthermore, it's an opt-in API, so by default nothing will change for existing callers.
1. Ensure that unit tests pass

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
+++ b/Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
@@ -98,11 +98,18 @@ extension URL {
         return url
     }
 
-    public func addParameter(name: String, value: String) throws -> URL {
+    public func addParameter(name: String, value: String, allowedReservedCharacters: CharacterSet? = nil) throws -> URL {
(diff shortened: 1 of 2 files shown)


tomasstrba (approved):
LGTM! 👍

PR #102, opened 2022-05-26
Add delegate to define if the credential should be automatically saved

Bunn:

**Required**:

Task/Issue URL: https://app.asana.com/0/414235014887631/1202348522197902/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1167
macOS PR:  https://github.com/more-duckduckgo-org/macos-browser/pull/598
What kind of version bump will this require?: Major, introduces a new delegate that needs to be implemented

**Description**:

Currently BSK assumes that if a credential has a password and not an username and the user is adding a username to the same domain with the same password, that we want this automatically saved.

Even though this works as intended for the macOS browser, on iOS we want to be notified when this scenario happens.
This PR adds a settings as delegate so the user can decide if they want the automatic save or not.

**Steps to test this PR**:
1.Run tests
1.Other than that, test on each plataforms own PRs should be enough

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


samsymons (approved):
Looks good :+1:

PR #106, opened 2022-06-07
Autofill logins iOS support

THISISDINOSAUR:

**Required**:

Task/Issue URL: 
iOS PR:  https://github.com/duckduckgo/iOS/pull/1182
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/606

**Optional**:

Tech Design URL:
CC:

**Description**:
Updates autofill to point at the new version with all the new autofill changes to support iOS, and makes various changes to e.g. AutofillUserScript to support logins on iOS.
Note, an actual version of the autofill JS script hasn't been released yet, so this is currently pointing at main. Therefore it probably shouldn't be tested until then, but the actual BSK code changes can be reviewed now.

**Steps to test this PR**:
1. Follow the testing steps in the macOS and iOS PRs to smoke test that nothing has changed or broken. Testing the actual new autofill functionality on iOS will come as a separate PR

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
```diff
@@ -19,6 +19,11 @@
 
 import WebKit
 
+public enum RequestVaultCredentialsAction: String, Codable {
```
brindy, line 22: Given `.none` case always accompanies nil credentials, and `.fill` is always associated with credentials, how about changing the handler so that it only returns an action and parameterise the fill case with the credentials in question, then credentials don't need to be optional either.

brindy (approved):
LGTM

PR #108, opened 2022-06-14
move removeParameter methods from iOS; replace parameter in addParameter

mallexxx:

**Required**:

Task/Issue URL: https://github.com/duckduckgo/iOS/issues/1191
iOS PR: https://github.com/duckduckgo/iOS/pull/1192
macOS PR: 

**Optional**:

Tech Design URL:
CC: @miasma13 

**Description**:
- Moved URL.removeParameter from iOS codebase and tests
- Updated addParameter to replace URL query parameter if exists

**Steps to test this PR**:
1. Validate addParameter works as expected
2. Validate iOS build works
3. Validate no API change required (minor version change)

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
+++ b/Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
@@ -115,6 +115,7 @@ extension URL {
         }
         
         var percentEncodedQueryItems = components.percentEncodedQueryItems ?? [URLQueryItem]()
+        percentEncodedQueryItems.removeAll(where: { $0.name == percentEncodedName })
         percentEncodedQueryItems.append(URLQueryItem(name: percentEncodedName, value: percentEncodedValue))
         components.percentEncodedQueryItems = percentEncodedQueryItems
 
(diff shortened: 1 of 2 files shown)


Sources/BrowserServicesKit/Common/Extensions/URLExtension.swift
```diff
@@ -121,6 +121,7 @@ extension URL {
         }
         
         var percentEncodedQueryItems = components.percentEncodedQueryItems ?? [URLQueryItem]()
+        percentEncodedQueryItems.removeAll(where: { $0.name == percentEncodedName })
```
miasma13, line 124: I suggest not to add this line that will cause replacing the parameter with the same name (is existing). Allowing multiple parameters with the same name is more aligned with the RFC specs and how `URLComponents` with `URLQueryItems` work.

Before removing that we also need to make sure if this won't be breaking change to iOS and need to review all the `addParameter` call use cases.

mallexxx: I‘ve checked again against our codebase and this change is based on the existing iOS code: [URLExtension.swift:L45](https://github.com/duckduckgo/iOS/blob/9e5d8fd3a6e1adcf6d434dda64aa8c8819d6b716/Core/URLExtension.swift#L45); [URLExtension.swift:L72](https://github.com/duckduckgo/iOS/blob/9e5d8fd3a6e1adcf6d434dda64aa8c8819d6b716/Core/URLExtension.swift#L72) which is removing all matching parameters when adding one
Also the function is only used for forming URLs for the search or for our API requests so this does not affect user browsing experience and URLs entered in the address bar - those are fixed separately

mallexxx:
I've update the URL parameters API naming to better describe their purpose, rolled back removing parameters and modified the API to allow passing same-named parameters in generic collection

miasma13 (approved):

PR #113, opened 2022-06-30
Elle/move secure vault to group

THISISDINOSAUR:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1202498365125439/f
iOS PR:  
macOS PR: 

**Optional**:

Tech Design URL:
CC:

**Description**:
This PR does three things:
- Moves secure vault on iOS to use an app group
- Enables write ahead logging
- Adds a check for autofill actually being enabled on the 'didRequestAutoFillInitDataForDomain" method, since if it's not, we shouldn't be trying to create a secure vault

The first two are shameless borrowed from Brindy's prototype to be a system level password provider (which is also why we want to make these changes, we want to do the app group change now so users don't lose all their data)

**Steps to test this PR**:
1. See macOS and iOS Pro for individual testing 

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/SecureVaultDatabaseProvider.swift
+++ b/Sources/BrowserServicesKit/SecureVault/SecureVaultDatabaseProvider.swift
@@ -68,6 +68,7 @@ final class DefaultDatabaseProvider: SecureVaultDatabaseProvider {
         var config = Configuration()
         config.prepareDatabase {
             try $0.usePassphrase(key)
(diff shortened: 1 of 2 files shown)


Bunn (approved):
There are some GHA warnings that I'm not sure if they were introduced on this PR, but other than that LGTM

PR #124, opened 2022-08-01
[WIP] Fire pixel when autoprompt is dismissed

GioSensation:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1202674424656875/f
iOS PR: WIP
macOS PR: WIP

**Description**:
This fires a pixel when autoprompt is dismissed.

**Steps to test this PR**:
**Pixel fired on iOS**
1. On iOS, visit a login page where you have credentials. For example, https://privacy-test-pages.glitch.me/autofill/autoprompt/case-1.html (save credentials for this page if you haven't yet).
2. Provided the form is front and center and not covered on pageload you should be prompted to autofill your credentials automatically (this is the case on the page linked above).
3. Dismiss the prompt and look at the Xcode console. You should see a pixel called `m_autofill_logins_autoprompt_dismissed`.

**No pixel on iOS**
1. On iOS, visit a login page where you have credentials but the form does not trigger autoprompt. For example, https://privacy-test-pages.glitch.me/autofill/autoprompt/case-2.html.2. 
2. Dismiss the prompt and look at the Xcode console. You should **not** see a pixel called `m_autofill_logins_autoprompt_dismissed`.

**No pixel on macOS**
1. On macOS, visit https://privacy-test-pages.glitch.me/autofill/autoprompt/case-1.html (you may have to save credentials for this page).
2. The autofill prompt should be open.
3. Dismiss the prompt by hitting Escape or clicking anywhere on the page.
4. You should **not** see the pixel `m_autofill_logins_autoprompt_dismissed`.

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
+++ b/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
@@ -41,6 +41,7 @@ public protocol AutofillSecureVaultDelegate: AnyObject {
                             completionHandler: @escaping ([SecureVaultModels.WebsiteAccount]) -> Void)
     func autofillUserScript(_: AutofillUserScript, didRequestCredentialsForDomain: String,
                             subType: AutofillUserScript.GetAutofillDataSubType,
+                            isAutoprompt: Bool,
                             completionHandler: @escaping (SecureVaultModels.WebsiteCredentials?, RequestVaultCredentialsAction) -> Void)
     
     func autofillUserScript(_: AutofillUserScript, didRequestCredentialsForAccount accountId: Int64,
@@ -329,6 +330,7 @@ extension AutofillUserScript {
     struct GetAutofillDataRequest: Codable {
         let mainType: GetAutofillDataMainType
         let subType: GetAutofillDataSubType
+        let isAutoprompt: Bool
     }
 
     // https://github.com/duckduckgo/duckduckgo-autofill/blob/main/src/deviceApiCalls/schemas/getAutofillData.params.json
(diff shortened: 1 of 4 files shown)


Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
```diff
@@ -329,6 +330,7 @@ extension AutofillUserScript {
     struct GetAutofillDataRequest: Codable {
         let mainType: GetAutofillDataMainType
         let subType: GetAutofillDataSubType
+        let isAutoprompt: Bool
```
shakyShane, line 333: @GioSensation could we represent this as an enum? firstly to match how it's represented in the JS side, but also to prevent restricting ourselves to a binary choice here.

samsymons (approved):
LGTM. Code looks fine and I've tested this via the client-side PRs.

GioSensation:
@samsymons Did you manage to look into the failing tests?

samsymons:
@GioSensation Yep, I've just pushed a commit to resolve the legit failure on this branch. There are some unrelated failures that happen only on macOS 12, we know about those and suspect they're an OS bug.

The issue with the legit failure was that the test suite was treating an asynchronous test as synchronous, and was completing too early. I've updated it to use a predicate expectation and to track which callbacks it receives – works better now. 🙂

GioSensation:
Awesome, thanks Sam! 🚀

PR #125, opened 2022-08-01
Remote Messaging

amddg44:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1201884213179453/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1263
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/709

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1202238116542877/f
CC:

**Description**:
This PR adds remote messaging functionality. See the task for more details

**Steps to test this PR**:
1. Testing steps are included in the client implementation PRs.
2. Run unit tests and ensure they are green

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [ ] macOS 10.15
* [ ] macOS 11

---
###### Internal references:


Sources/BrowserServicesKit/RemoteMessaging/Model/AnyDecodable.swift
```diff
@@ -0,0 +1,183 @@
...
+// portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+//
+// https://github.com/Flight-School/AnyCodable/blob/master/Sources/AnyCodable/AnyDecodable.swift
```
bwaresiak, line 19: Question: why this is embedded as a file instead of a dependency?

amddg44: The library is not regularly maintained and I only needed one file. I can include as a dependency but I made a judgement call. Happy to add this as a dependency if you prefer?

Sources/BrowserServicesKit/Statistics/VariantManager.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import Foundation
+
+public enum FeatureName: String {
+
+    // Used for unit tests
+    case dummy
+}
+
+public protocol VariantManager {
```
THISISDINOSAUR, line 28: There's a few things like this and the statistics store protocol that are moved from iOS, right? 
I'm good with that since I think in all the cases it's stuff that macOS could benefit from at some point the future, but did want to suggest an alternative for the future: typically if there's info something in BSK needs from the client we usually have some kind of delegate back to the client that returns the info. Like everything else it's not really an official policy or decision, but I think it's become a thing implicitly

Re StatisticsStore it looks like we now have a different version of in all three repos, so in that case I'd either change it to use the above pattern, or better yet and if it makes sense to share it, change iOS and macOS to both use the one now in BSK

THISISDINOSAUR (commented):
I'm struggling to wrap my head around some of the json parsing, can we chat about it on Monday?

Sources/BrowserServicesKit/RemoteMessaging/Model/AnyDecodable.swift
bwaresiak: I think this is fine as long as we make sure that license is embedded with the file license is compatible with ours - both seem to be the case here.

Sources/BrowserServicesKit/Common/Extensions/CalendarExtension.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+
+import Foundation
+
+extension Calendar {
+	func numberOfDaysBetween(_ from: Date, and to: Date) -> Int? {
+		let numberOfDays = dateComponents([.day], from: from, to: to)
+		return numberOfDays.day
+	}
+}
```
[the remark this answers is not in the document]
bwaresiak: Yea, also if added, things like this could benefit from being a module. Not saying to do that now but we are long overdue with discussion about it.

bwaresiak:
I'm assigning myself as secondary reviewer as this is pretty big work in shared space (BSK). Will take a look on this Tomorrow (Tue).

Sources/BrowserServicesKit/Common/InfoBundle.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public protocol InfoBundle {
```
THISISDINOSAUR, line 22: Does this protocol exist just for testing purposes? (MockBundle was the only usage I could find)

Sources/BrowserServicesKit/Domain/AppVersion.swift
```diff
@@ -0,0 +1,74 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+/// 1. App version is moved to BSK and it's assumed Mac can use this struct without any changes.
+/// 2. Create AppVersion protocol in BSK and create struct that implements it in Core app called iOSAppVersion.swift. (APMobile)
```
THISISDINOSAUR, line 23: I'm not sure I understand this comment

Sources/BrowserServicesKit/Domain/AppVersion.swift
```diff
@@ -0,0 +1,74 @@
...
+
+public struct AppVersion {
```
THISISDINOSAUR, line 25: Regardless of pending conversations wrt BSK I'm not sure I like this being in BSK since it's info so inherent to the client. But not sure I can substantiate that feeling, so if you are reading this Bartek I'm keen to see what you think

Sources/BrowserServicesKit/RemoteMessaging/Mappers/JsonRemoteMessageMapper.swift
```diff
@@ -0,0 +1,211 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import os.log
+
+private enum AttributesKey: CaseIterable {
+    static let locale             = "locale"
+    static let osApi              = "osApi"
+    static let flavor             = "flavor"
```
THISISDINOSAUR, line 26: This isn't really a meaningful name to me, is it standard language in android land?
If so, I think this would be a good point to deviate from them since it's not in apple land

Sources/BrowserServicesKit/RemoteMessaging/Mappers/JsonRemoteMessageMapper.swift
```diff
@@ -0,0 +1,211 @@
...
+            return .ddgAnnounce
+        case .criticalUpdate:
+            return .criticalUpdate
+        case .none:
+            return .announce
+        }
+    }
+
+    // swiftlint:disable cyclomatic_complexity
+    static func maps(jsonRemoteRules: [RemoteMessageResponse.JsonMatchingRule]) -> [Int: [MatchingAttribute]] {
```
THISISDINOSAUR, line 152: This function feels like a lot of boilerplate to me that I'n not sure is necessary, especially as for each case it does a very similar call out to JsonRulesMapper, which then seems to do the same boiler plate in each case

Sources/BrowserServicesKit/RemoteMessaging/Matchers/AppAttributeMatcher.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    private let bundleId: String
+    private let appVersion: String
+    private let flavor: AppFlavor
+    private let statisticsStore: StatisticsStore
+    private let variantManager: VariantManager
+
+    public init(statisticsStore: StatisticsStore, variantManager: VariantManager, flavor: AppFlavor = .flavorPublic) {
+        if let bundleId = Bundle.main.bundleIdentifier {
+            self.init(bundleId: bundleId,
```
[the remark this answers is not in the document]
THISISDINOSAUR: Although in what cases will it end up as ""? How hard would it be to make it optional? I could see the use of an empty string getting confusing down the line

Sources/BrowserServicesKit/RemoteMessaging/Model/AnyDecodable.swift
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+@usableFromInline
+protocol _AnyDecodable {
+    var value: Any { get }
+    init<T>(_ value: T?)
+}
+
+extension AnyDecodable: _AnyDecodable {}
```
THISISDINOSAUR, line 39: What is this file used for?

Sources/BrowserServicesKit/RemoteMessaging/RemoteMessagingConfigProcessor.swift
```diff
@@ -0,0 +1,63 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import os.log
+
+public struct RemoteMessagingConfigProcessor {
```
THISISDINOSAUR, line 23: Would it be possible to rename this? Atm it's not immediately clear to me where it sits in the pipeline

THISISDINOSAUR:
I think it is worth us discussing over zoom, I'll need to be convinced this JSON parsing strategy is the way to go 😅

Sources/BrowserServicesKit/Common/Extensions/CalendarExtension.swift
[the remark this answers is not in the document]
brindy: Isn't this internal to the package anyway?  

I'd be more worried* if it were public, but I don't think we should be overly restrictive about such things if it makes the code in this package more readable, etc.

\* I'm not worried at all

Sources/BrowserServicesKit/Common/InfoBundle.swift
brindy: I believe it does, and was pulled from iOS?  

I was tackling this myself when working on sync before I moved to that other project.  I'm not keen on it, but the only way to create tests for the Bundle related stuff is to create fake bundles.  I'm think I'd prefer to remove this if possible.

Sources/BrowserServicesKit/Domain/AppVersion.swift
brindy: Personally, I think it's fine. This library is intended to support making development of the clients easier/quicker and our clients have these keys so it seems reasonable to share this code imo.  

fwiw I had also started to do this kind of thing in the sync branches before I got seconded.  What I want to avoid is spaghetti code and creating additional protocols and wiring for the clients is just that - hence, I also support moving the network layer into this package as well, but I guess we'll want to discuss all that more widely.

Sources/BrowserServicesKit/RemoteMessaging/Matchers/AppAttributeMatcher.swift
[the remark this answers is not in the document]
brindy: I would as in what case do we expect the main bundle not to have a bundle identifier and actually force unwrap it, or at least have an `assert`.

Sources/BrowserServicesKit/RemoteMessaging/Mappers/JsonRemoteMessageMapper.swift
brindy: Yep, it is. Check out `AppFlavors` and this https://github.com/duckduckgo/Android/blob/develop/app/build.gradle#L76

I agree, it has no place here and should be removed. I don't know if that means having a different JSON file for Apple like we do with the privacy config or what, but that needs to be resolved.

THISISDINOSAUR: The plan is to have separate JSON files for both, not sure if the specs are currently entirely identical though?

Sources/BrowserServicesKit/RemoteMessaging/Model/AnyDecodable.swift
brindy: @amddg44 can correct me if I'm wrong, but I believe this facilitates parsing of dictionaries like the `translations` while still using a `JSONDecoder()` 

See https://github.com/duckduckgo/BrowserServicesKit/pull/125/files#diff-55a677ad8562e792bf8f1384c1f39effbf7947777399f208653996959d2f0bb5R17

Would have been better imo if it were a array of translation objects with the local and as attribute, then map that to a dictionary in memory after it was parsed, but don't think we can fix that that now.

Sources/BrowserServicesKit/Statistics/VariantManager.swift
brindy: >  typically if there's info something in BSK needs from the client we usually have some kind of delegate back to the client that returns the info.

I said it elsewhere, but actually I disagree with this as we just end up creating a spaghetti architecture. Obviously not the place to debate it so let's get it on the team agenda soon.

Sources/BrowserServicesKit/Common/InfoBundle.swift
amddg44: InfoBundle is required by AppVersion which I also moved to BSK

Sources/BrowserServicesKit/Domain/AppVersion.swift
amddg44: This was an earlier comment I forgot to remove, sorry for the confusion

Sources/BrowserServicesKit/Domain/AppVersion.swift
amddg44: AppVersion is required for the remote messaging rules so it is possible of course to cherry pick the properties needed but I felt like this was the best approach.

Remote messaging framework uses AppVersion like the following:
- if AppVersion is X, do Y
- if AppVersion is > X, but < Y, do Z 
- etc

Sources/BrowserServicesKit/RemoteMessaging/RemoteMessagingConfigProcessor.swift
amddg44: Absolutely, as discussed lets talk with Cristian when he returns but no problem to rename

Sources/BrowserServicesKit/RemoteMessaging/Model/AnyDecodable.swift
amddg44: The reason for AnyDecodable is primarily based upon decoding the rules with its various structures. I guess the other approach is to change the iOS json file structure in a way that AnyDecodable wouldn't be needed - however this could potentially create a significant amount of work for whoever is creating the json files. The process would no longer be start with one platform, then copy for second platform and do a couple of platform specific tweaks.

https://github.com/duckduckgo/BrowserServicesKit/blob/0c0b8114f8121ecf0886d7ec2f7f2e11fc3f683a/Tests/BrowserServicesKitTests/Resources/remote-messaging-config.json#L102

Sources/BrowserServicesKit/Common/InfoBundle.swift
THISISDINOSAUR: Okay, if this already existed I'm fine with it

THISISDINOSAUR:
Thinking about it after our chat and having had a brief look at the android code, I think doing the parsing/processing in the same way as android has led to something a bit less elegant here.

I understand the reasoning for keeping the JSON specs the same, but I'm not convinced keeping the structure of the processing the same is worthwhile

I'd suggest one of these two options:
1.
- Remove some of the repetition, e.g. like here https://github.com/duckduckgo/BrowserServicesKit/pull/127/files
- Rename some things like we talked about it to make it clearer what each step is doing
- Add some documentation somewhere to aid the above

2. 
- A more substantial change to completely disregard what android is doing

Sources/BrowserServicesKit/RemoteMessaging/Mappers/JsonRemoteMessageMapper.swift
amddg44: I've changed the property and it's type to be the same as FeatureFlaggerInternalUserDecider.isInternalUser

Sources/BrowserServicesKit/RemoteMessaging/Matchers/AppAttributeMatcher.swift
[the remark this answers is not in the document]
amddg44: I've updated this to re-use `AppVersion.shared.identifier` which was already doing the call and added an assert in case this is ever empty.

amddg44:
> I'd suggest one of these two options: 1.
> 
> * Remove some of the repetition, e.g. like here https://github.com/duckduckgo/BrowserServicesKit/pull/127/files

LGTM, I've pushed a change using your example

THISISDINOSAUR (approved):
Great, LGTM now!

bwaresiak (approved):
LGTM! :)

PR #135, opened 2022-08-29
Try/catch on contentblocking message handler

SlayterDev:

**Required**:

Task/Issue URL: https://app.asana.com/0/1200223097357040/1202818860070691/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1291
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/704

**Optional**:

Tech Design URL:
CC: @jonathanKingston @bwaresiak 

**Description**:
This PR will fallback to `window.webkit` if the global variable isn't found. This seems to be a case that can happen for various reasons as of iOS 14.2. https://stackoverflow.com/questions/64922289/cant-find-variable-webkit-on-ios-14-2

**Steps to test this PR**:
1. Visit the site listed in the Asana task.
1. Site should load. (Previously a blank page)

**OS Testing**:

* [x] iOS 13
* [x] iOS 14
* [x] iOS 15
* [x] macOS 10.15
* [x] macOS 11

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
+++ b/Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
@@ -170,14 +170,14 @@
         })
     }
 
-    if (webkit.messageHandlers.processRule) {
+    if ((webkit || window.webkit).messageHandlers.processRule) {
         install()
     }
 
     function install () {
         function sendMessage (url, resourceType) {
             if (url) {
-                webkit.messageHandlers.processRule.postMessage({
+                (webkit || window.webkit).messageHandlers.processRule.postMessage({
                     url: url,
                     resourceType: resourceType === undefined ? null : resourceType,
                     blocked: !unprotectedDomain && !isTrackerAllowlisted(topLevelUrl, url),
--- a/Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
+++ b/Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
@@ -177,12 +177,14 @@
     function install () {
         function sendMessage (url, resourceType) {
             if (url) {
-                (webkit || window.webkit).messageHandlers.processRule.postMessage({
-                    url: url,
-                    resourceType: resourceType === undefined ? null : resourceType,
-                    blocked: !unprotectedDomain && !isTrackerAllowlisted(topLevelUrl, url),
-                    pageUrl: topLevelUrl.href
-                })
+                try {
+                    (webkit || window.webkit).messageHandlers.processRule.postMessage({
+                        url: url,
+                        resourceType: resourceType === undefined ? null : resourceType,
+                        blocked: !unprotectedDomain && !isTrackerAllowlisted(topLevelUrl, url),
+                        pageUrl: topLevelUrl.href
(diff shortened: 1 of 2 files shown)


Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
```diff
@@ -170,19 +170,21 @@
         })
     }
 
-    if (webkit.messageHandlers.processRule) {
+    if ((webkit || window.webkit).messageHandlers.processRule) {
```
jonathanKingston, line 173: These are equivalent, I don't this will make any difference.

Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
```diff
@@ -170,19 +170,21 @@
...
         function sendMessage (url, resourceType) {
             if (url) {
-                webkit.messageHandlers.processRule.postMessage({
-                    url: url,
-                    resourceType: resourceType === undefined ? null : resourceType,
-                    blocked: !unprotectedDomain && !isTrackerAllowlisted(topLevelUrl, url),
-                    pageUrl: topLevelUrl.href
-                })
+                try {
+                    (webkit || window.webkit).messageHandlers.processRule.postMessage({
```
jonathanKingston, line 181: Perhaps the try catch is the only difference here.

As mentioned perhaps there's an execution context that doesn't have the message handlers available.

Sources/BrowserServicesKit/ContentBlocking/UserScripts/contentblockerrules.js
SlayterDev: I was a bit curious about this. And it actually did on my first try! (I swear I'm not making it up) but as you mentioned in the other comment the try/catch is the real change. I can remove the fallback.

jonathanKingston: I wonder if it's just a race condition of contexts loading then?

It would be good to work out what handlers we're expecting and have enabled still. As you also mentioned perhaps some have been removed by mistake.

SlayterDev: In this case its not the handler that doesn't exist. Its the entire `webkit` object that is missing. On my first run the fallback to `window.webkit` fixed the issue which is documented as a divergence in iOS versions (allegedly) [here](https://stackoverflow.com/questions/64922289/cant-find-variable-webkit-on-ios-14-2). But after verifying and applying this here it no longer worked until I put in the try/catch (for obvious reasons).

Sources/BrowserServicesKit/ContentBlocking/UserScripts/surrogates.js
```diff
@@ -21,7 +21,7 @@
     const duckduckgoDebugMessaging = (function () {
         let log = () => {}
         let signpostEvent = () => {}
-
+
```
jonathanKingston, line 24: Can you run `git checkout origin/main -- Sources/BrowserServicesKit/ContentBlocking/UserScripts/surrogates.js` or similar?

jonathanKingston (approved):
🚀

PR #148, opened 2022-10-06
Add script execution guard for about:blank frames

jonathanKingston:

**Required**:

Task/Issue URL: https://app.asana.com/0/1200437802575119/1202155258470004/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1335
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/750

**Optional**:

Tech Design URL:
CC:

**Description**:

Visiting https://www.washingtonpost.com/ without this patch you'll see "TypeError: Attempting to change value of a readonly property." errors as we're creating a property called __firefox__.find (in the isolated world) and then trying to redeclare over it.

This is only a side effect of us executing all user scripts multiple times for about:blank frames; most of our code isn't idempotent so will be creating side effects.

The solution is a little hacky here but we create a guard that prevents the script from executing multiple times for the same execution context.

- Note when debugging this; we run some scripts in isolated and some in default worlds so the variable is created both times. The issue exists in both worlds but the array is created both times so you'll see different sized arrays as you step through etc (this is because isolated worlds have a totally different window global) the code works as expected for the isolated also.

**Steps to test this PR**:
1. Visiting https://www.washingtonpost.com/
2. There should be no error with the following text in the console: "TypeError: Attempting to change value of a readonly property."

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/UserScript/UserScript.swift
+++ b/Sources/BrowserServicesKit/UserScript/UserScript.swift
@@ -18,6 +18,7 @@
 //
 
 import WebKit
+import CryptoKit
 
 public protocol UserScript: WKScriptMessageHandler {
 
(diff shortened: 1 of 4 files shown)


Sources/BrowserServicesKit/UserScript/UserScript.swift
```diff
@@ -73,11 +74,23 @@ extension UserScript {
     static func makeWKUserScript(source: String, injectionTime: WKUserScriptInjectionTime,
                                  forMainFrameOnly: Bool,
                                  requiresRunInPageContentWorld: Bool = false) -> WKUserScript {
+        let hash = SHA256.hash(data: Data(source.utf8)).hashValue
+
+        // This prevents the script being executed twice which appears to be a WKWebKit issue for about:blank frames when the location changes
+        let sourceOut = """
+        (() => {
+            if (window.navigator._duckduckgoloader_ && window.navigator._duckduckgoloader_.includes('\(hash)')) {return}
```
shakyShane, line 82: can we make this check `Array.isArray()` instead - just to ensure `.includes()` is valid

jonathanKingston: What's the rationale?

I'm only checking the presence as it's not defined first time round, the definition is always within the same file.

I don't think we should be defending against the page if that's your intent here; we'd need to take prototypes and prevent overriding etc.

jonathanKingston: Resolved over MM :)

samsymons (approved):
This looks good as far as I can see, let's give it a try. 🙂 

* I ran through JS user script testing in various forms to verify that everything is working as expected
* Confirmed that the WaPo issue is gone
* Checked the web inspector console for any other JS issues that seemed suspicious - nothing came up
* Added logging to the new JS to confirm that the early return based on hash lookup is being triggered successfully

PR #157, opened 2022-10-27
Checking whether domains of credential and requesting website mach

tomasstrba:

**Required**:

Task/Issue URL: https://app.asana.com/0/1177771139624306/1203127424124512/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1362
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/803

**Description**:

This PR adds a security check that injects the credentials into the website only if domains match

**Steps to test this PR**:
1. Make sure autofill works for multiple items from your keychain
1. Verify you no longer can request the credentials for an arbitrary id and get it injected into a page.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
+++ b/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
@@ -439,7 +439,10 @@ extension AutofillUserScript {
             guard let credential = $0,
                   let id = credential.account.id,
                   let password = String(data: credential.password, encoding: .utf8),
-                  credential.account.domain == requestingDomain else { return }
+                  credential.account.domain == requestingDomain else {
+                replyHandler("{}")
+                return
+            }
 
             let response = RequestVaultCredentialsForAccountResponse(success: .init(id: String(id),
                                                                     username: credential.account.username,
--- a/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
+++ b/Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
@@ -430,11 +430,13 @@ extension AutofillUserScript {
               let accountId = Int64(id) else {
             return
         }
+        let requestingDomain = hostForMessage(message)
 
         vaultDelegate?.autofillUserScript(self, didRequestCredentialsForAccount: Int64(accountId)) {
             guard let credential = $0,
                   let id = credential.account.id,
-                  let password = String(data: credential.password, encoding: .utf8) else { return }
+                  let password = String(data: credential.password, encoding: .utf8),
+                  credential.account.domain == requestingDomain else { return }
 
             let response = RequestVaultCredentialsForAccountResponse(success: .init(id: String(id),
                                                                     username: credential.account.username,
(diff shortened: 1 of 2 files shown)


Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
```diff
@@ -430,11 +430,13 @@ extension AutofillUserScript {
...
             return
         }
+        let requestingDomain = hostForMessage(message)
 
         vaultDelegate?.autofillUserScript(self, didRequestCredentialsForAccount: Int64(accountId)) {
             guard let credential = $0,
                   let id = credential.account.id,
-                  let password = String(data: credential.password, encoding: .utf8) else { return }
+                  let password = String(data: credential.password, encoding: .utf8),
+                  credential.account.domain == requestingDomain else { return }
```
tomasstrba, line 439: @GioSensation, do you have any idea why don't send a response in case of failure here? Is it possible to send failure in _RequestVaultCredentialsForAccountResponse_?

tomasstrba (commented):
There is one failing unit test caused by the fact we only send the answer in case of success.

Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
GioSensation: We don't have an error UI or flow. These cases should not be possible in normal usage because if we're here it means we already know credentials exist for this domain (we inject the keys/listeners only if credentials exist and show a credentials list in the tooltip picker). We technically have both `success` and `error` objects in the response but no platform has ever implemented anything with `error` in there, again because there is no UI.

What we should do here is probably send an error pixel because this shouldn't be possible.

tomasstrba: 👍 , adjusted the unit test so there is not expectation for an answer in case of error

tomasstrba:
Checks passing, ready for a review 👍

Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
GioSensation: Can we do 
```suggestion
                  credential.account.domain == requestingDomain else {
                      // Fire error pixel here if possible
                      replyHandler("{}")
                      return   
                  }
```

GioSensation (changes requested):
@tomasstrba It's better if we reply with an empty object. It enters a catch block on our end. Again, this shouldn't be possible but it's cleaner this way. Also, if we have have an error pixel to fire that'd be grand, but no biggie otherwise.

tomasstrba:
@GioSensation, I implemented changes you proposed and fix SwiftLint errors. There is no common mechanism to fire pixel from BSK, so I didn't invest time into it but we can scope a follow up to resolve it in general.
The PR is ready for a final review

GioSensation (approved):
Thanks, Tom! 🚀

PR #234, opened 2023-02-20
WKFrameInfo nullable request property

mallexxx:

**Required**:

Task/Issue URL: https://app.asana.com/0/1199230911884351/1203965979591356/f
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/981
What kind of version bump will this require?: Minor/Patch

CC: @tomasstrba 

**Description**:
Added WKFrameInfo.safeRequest property instead of wrongly-non-nullable .request
Added runtime checks for WKFrameInfo.safeRequest and WKNavigationAction.sourceFrame usage in Debug config

Steps to test this PR:
0. Validate both Debug&Release builds
1. Go to https://www.seriouseats.com/ask-the-food-lab-do-i-need-to-use-kosher-salt
2. Click through the "Buy on ..." product links, validate there‘s no crash when WKFrameInfo.request is accessed
3. Validate microphone permission works on https://permission.site/

### DIFF
--- a/Sources/Navigation/Extensions/WKFrameInfoExtension.swift
+++ b/Sources/Navigation/Extensions/WKFrameInfoExtension.swift
@@ -1,5 +1,5 @@
 //
-//  WKNavigationActionExtension.swift
+//  WKFrameInfoExtension.swift
 //
 //  Copyright © 2023 DuckDuckGo. All rights reserved.
 //
--- a/Sources/Navigation/Extensions/WKFrameInfoExtension.swift
+++ b/Sources/Navigation/Extensions/WKFrameInfoExtension.swift
@@ -1,0 +1,110 @@
+//
+//  WKNavigationActionExtension.swift
+//
+//  Copyright © 2023 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


Sources/Navigation/Extensions/WKFrameInfoExtension.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Safe Optional `request: URLRequest` getter:
+    /// .request of a new Frame can be `null`, see https://app.asana.com/0/0/1203965979591356/f
+    var safeRequest: URLRequest? {
+        _=WKFrameInfo.addSafetyCheckForSafeRequestUsageOnce
+        return self.perform(#selector(getter: request))?.takeUnretainedValue() as? URLRequest
+    }
+
+#if DEBUG
+    private static var ignoredRequestUsageSymbols = Set<String>()
```
ayoy, line 48: Definitely not blocking this PR, but given we use it in 3 places, we may want to abstract this mechanism for future use at some point of time (QWD?).

ayoy (approved):
LGTM! Crashing is gone (I've also checked with the release build), thanks for handling that so quickly @mallexxx!

Sources/Navigation/Extensions/WKFrameInfoExtension.swift
mallexxx: maybe in future, given that it‘s not business code duplication I‘ll live with it for now

PR #247, opened 2023-03-02
fix assertion for popup about:blank navigations without WKNavigation

mallexxx:

**Required**:

Task/Issue URL: https://app.asana.com/0/1177771139624306/1204065038916789/f
iOS PR:  - not affected
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1008

**Optional**:

**Steps to test this PR**:
1. window.open("") from js console, validate no assertion is raised

### DIFF
--- a/Sources/Navigation/DistributedNavigationDelegate.swift
+++ b/Sources/Navigation/DistributedNavigationDelegate.swift
@@ -413,7 +413,11 @@ extension DistributedNavigationDelegate: WKNavigationDelegate {
     @MainActor
     public func webView(_ webView: WKWebView, didStartProvisionalNavigation wkNavigation: WKNavigation?) {
         let navigation: Navigation
-        if let expectedNavigation = navigationExpectedToStart, wkNavigation != nil || expectedNavigation.navigationAction.navigationType == .sessionRestoration {
+        if let expectedNavigation = navigationExpectedToStart,
+           wkNavigation != nil
+            || expectedNavigation.navigationAction.navigationType == .sessionRestoration
+            || expectedNavigation.navigationAction.url.scheme.map(URL.NavigationalScheme.init) == .about {
+
             // regular flow: start .expected navigation
             navigation = expectedNavigation
         } else if webView.url?.isEmpty == false {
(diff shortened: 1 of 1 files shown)


Sources/Navigation/DistributedNavigationDelegate.swift
```diff
@@ -696,7 +698,7 @@ extension DistributedNavigationDelegate: WKNavigationDelegate {
     @MainActor
     public func webView(_ webView: WKWebView, didCommit wkNavigation: WKNavigation?) {
         guard let navigation = wkNavigation?.navigation ?? startedNavigation else {
-            assertionFailure("Unexpected didCommitNavigation")
+            assert(wkNavigation == nil, "Unexpected didCommitNavigation without preceding didStart")
```
ayoy, line 701: Kudos for the more descriptive assertion message!

Sources/Navigation/DistributedNavigationDelegate.swift
```diff
@@ -416,10 +416,12 @@ extension DistributedNavigationDelegate: WKNavigationDelegate {
         if let expectedNavigation = navigationExpectedToStart, wkNavigation != nil || expectedNavigation.navigationAction.navigationType == .sessionRestoration {
             // regular flow: start .expected navigation
             navigation = expectedNavigation
-        } else {
+        } else if webView.url?.isEmpty == false {
```
ayoy, line 419: While it fixes calling `window.open(“”)`, it doesn’t seem to handle `window.open(“about:settings”)` which is used in Duck Player to open settings and as such causes an assertion on debug builds.

ayoy (changes requested):
Thanks for the fix @mallexxx! I’ve found that even with this patch, opening settings from Duck Player (via ⚙️ icon) crashes on the assertion. Please have a look when you’re back. Thanks again.

Sources/Navigation/DistributedNavigationDelegate.swift
mallexxx: fixed

PR #248, opened 2023-03-02
Add xcbeautify to unit tests reports

jotaemepereira:
# Task issue

https://app.asana.com/0/1203301625297703/1203998658300743/f

## Description

Add [xcbeautify](https://github.com/tuist/xcbeautify) to our unit tests report.

I also removed the need to add the --parellel flag because now we use the xcbeautify report instead of the swift test one

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -44,13 +44,12 @@ jobs:
     - name: Select Xcode
       run: sudo xcode-select -s /Applications/Xcode_14.1.app/Contents/Developer
 
+    - name: Install xcbeautify
+      continue-on-error: true
+      run: brew install xcbeautify
+
     - name: Run tests
-      # BSK does not support running tests in parallel, but the flag is required
-      # to make XUnit output work. Making it parallel with 1 worker then.
-      # https://stackoverflow.com/a/70040836
-      run: |
-        swift test --parallel --num-workers=1 \
-          --xunit-output tests.xml
+      run: swift test | xcbeautify --report junit --report-path . --junit-report-filename tests.xml
 
     - name: Publish Unit Tests Report
       uses: mikepenz/action-junit-report@v3
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -49,7 +49,7 @@ jobs:
       run: brew install xcbeautify
 
     - name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -44,13 +44,12 @@ jobs:
...
+      run: brew install xcbeautify
+
     - name: Run tests
-      # BSK does not support running tests in parallel, but the flag is required
-      # to make XUnit output work. Making it parallel with 1 worker then.
-      # https://stackoverflow.com/a/70040836
-      run: |
-        swift test --parallel --num-workers=1 \
-          --xunit-output tests.xml
+      run: swift test | xcbeautify --report junit --report-path . --junit-report-filename tests.xml
```
ayoy, line 52: I’ve pushed a commit that breaks tests and it still managed to report success.
This may be because xcbeautify overrifing the non-zero exit status code of swift test.
To propagate status code down the Unix pipe, use pipefail modifier such as:

```bash
set -o pipefail && swift test | xcbeautify ...
```

ayoy (changes requested):
Thanks for this @jotaemepereira! Kudos for removing the `--num-workers=1` argument as we’ve actually fixed parallel unit testing some time ago and never fixed the CI :) that should be a significant speed up :)

I’ve found that after your changes the workflow doesn’t report failing unit tests. Please have a look at my comment and try with a pipefail modifier. I suppose it may fix the issue. Thanks again!

.github/workflows/unit-tests.yml
jotaemepereira: Thank you for spotting that 👏🏼 ! I only tested failed tests and success, but I didn't test if tests didn't compile.

I will add the command to propagate the status code, and I will test the different scenarios and let you know

jotaemepereira:
@ayoy This should be fixed. I've tested the following scenarios:

- If tests succeeded, it will show the ✅ check for the Test Report with information about how many tests ran
- If some tests fail, it will show the ❌ for both the Run unit tests flow and Test Report
- If tests fail to compile, it will show the ❌ for the Run unit tests, but it will show the ✅ for Test Report. This is because the `Publish Unit Tests Report` always runs even if running the steps failed. So I've added the `require_tests` so the report fails if there are no tests available.

I think it makes sense to fail the test report if no unit tests were found. What do you think?

ayoy (approved):
LGTM! I’ve checked your test  workflow runs and all looks great now. Thanks a lot @jotaemepereira 🎉

PR #256, opened 2023-03-14
HTTPS upgrade threading

mallexxx:

**Required**:

Task/Issue URL: https://app.asana.com/0/1177771139624306/1204105120334404/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1536
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1031

**Description**:
- Updating threading for Bloom Filter loading

**Steps to test this PR**:
- Validate threading correctness
- Validate config updating bloom filter

### DIFF
--- a/Sources/BrowserServicesKit/SmarterEncryption/HTTPSUpgrade.swift
+++ b/Sources/BrowserServicesKit/SmarterEncryption/HTTPSUpgrade.swift
@@ -26,7 +26,7 @@ public enum HTTPSUpgradeError: Error {
     case nonHttp
     case domainExcluded
     case featureDisabled
-    case nonUpgradable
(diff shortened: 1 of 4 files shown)


brindy (approved):
LGTM and I've approved macOS and iOS too.

I tested both platforms. Seems to work fine, don't see any errors when upgrading.

What I notice is a there's a LOT of commonality now that seems to be copied between both platforms.

Happy for you to just go ahead and merge this, but I wonder if we should move that common code into BSK?  For example `AppHTTPSUpgradeStore` on both platforms look the same now, I think?  imo that could replace the protocol in BSK with this concrete class - or something.

Anyway, something we should consider as a separate PR!  I'll create a task for it.

PR #258, opened 2023-03-14
Add PrivacyConfig subfeatures + remote internal user feature flagging

graeme:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1203676877763836
iOS PR:  
macOS PR: 

**Optional**:

Tech Design URL:
https://app.asana.com/0/481882893211075/1203938569856702
https://app.asana.com/0/481882893211075/1203984214904257
CC: 

**Description**:

This PR adds subfeatures support to the Privacy Config component, adds support for an additional state, internal, and moves the InternalUserDecider from iOS into BSK to be used by the `AppPrivacyConfiguration` to provide the correct state for internal users.

I'm mainly opening this as a draft to get some initial eyes on it before opening a final PR. It needs more testing (automated and manual). I'm going to be separating the subfeature part out in parallel as this is more pressing for the Autofill project.

**Steps to test this PR**:
1. Check out the iOS PR above
2. Check that any existing features configured using the privacy-config still work as expected

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/BrowserServicesKit/InternalUserDecider/InternalUserDecider.swift
```diff
@@ -0,0 +1,78 @@
...
+    @discardableResult
+    func markUserAsInternalIfNeeded(forUrl url: URL?, response: HTTPURLResponse?) -> Bool
+}
+
+public protocol InternalUserStore {
+    var didVerifyInternalUser: Bool { get set }
+}
+
+public class DefaultInternalUserDecider: InternalUserDecider {
+    private let userDefaults = UserDefaults()
```
graeme, line 35: This may be controversial as I can only see one other place in BSK that manages UserDefaults access itself and not through an injected store. This change would also meant that everyone would have to re-authenticate which would be mildly inconvenient for internal users only.

Sources/BrowserServicesKit/InternalUserDecider/InternalUserDecider.swift
```diff
@@ -0,0 +1,78 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public protocol InternalUserDecider {
+    
+    var isInternalUser: Bool { get }
```
graeme, line 24: This could be all we use to flag internal features, replacing the feature by feature flagging (at least while in development).

Sources/BrowserServicesKit/InternalUserDecider/InternalUserDecider.swift
graeme: Actually need to inject a store as it's encrypted on macOS. Shouldn't be too hard

THISISDINOSAUR: At any rate, I would definitely rather avoid people having to reauthenticate if possible, we have a good adoption rate at the moment, and I'm not sure how easy that will be to replicate.

Sources/BrowserServicesKit/InternalUserDecider/InternalUserDecider.swift
```diff
@@ -0,0 +1,78 @@
...
+        }
+        
+        if shouldMarkUserAsInternal(forUrl: url, statusCode: response?.statusCode) {
+            didVerifyInternalUser = true
+            return true
+        }
+        return false
+    }
+    
+    func shouldMarkUserAsInternal(forUrl url: URL?, statusCode: Int?) -> Bool {
```
THISISDINOSAUR, line 68: When you factor out the storage, I might be tempted to find a way to make this shared, since I can't imagine the rules being different for each platform

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -76,10 +83,28 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+        guard let feature = data.features[featureKey.key] else { return false }
+        let satisfiesMinVersion = satisfiesMinVersion(feature.minSupportedVersion, versionProvider: versionProvider)
+        switch feature.state {
+        case PrivacyConfigurationData.State.enabled: return satisfiesMinVersion
+        case PrivacyConfigurationData.State.internal: return internalUserDecider.isInternalUser && satisfiesMinVersion
+        default: return false
+        }
+    }
+
+    public func isSubfeatureEnabled<F: NestedFeature, SF>(for feature: F.Type, _ subfeature: SF, versionProvider: AppVersionProvider) -> Bool where F.SubFeatureType == SF {
```
THISISDINOSAUR, line 95: If the parent feature is set to disabled, but the subfeature is set to enabled, should the sub feature resolve to enabled?
I think it probably shouldn't be, did we get alignment on that?

Sources/BrowserServicesKit/FeatureFlagger/FeatureFlagger.swift
```diff
@@ -0,0 +1,61 @@
...
+        switch provider.source {
+        case .disabled:
+            return false
+        case .internalOnly:
+            return internalUserDecider.isInternalUser
+        case .remoteDevelopment(let configHandler):
+            guard internalUserDecider.isInternalUser else {
+                return false
+            }
+            fallthrough
```
THISISDINOSAUR, line 56: This is the first instance of fallthrough in either the iOS or BSK codebases. We don't have a collective agreement to not use it or anything, although I would personally rather avoid it unless it's avoiding much more repetition.

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
graeme: You're right, good catch. This was an oversight but it's a simple a fix as just guarding the parent feature's value first.

PR #281, opened 2023-03-28
sync conect

brindy:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1204266236802307/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/1604
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1062

**Optional**:

Tech Design URL:
CC:

**Description**:

Add sync connect flow.

**Steps to test this PR**:

***iOS connect from macOS***

1. Create an account on an iOS device or simulator
2. Launch macOS, remove existing account if needed
3. Begin turn on sync flow and continue to sync with another device - a QR code should be visible
4. On iOS tap on Scan QR Code 
5. Using Manually Enter Code paste in the code
6. Wait a few seconds and the account should appear on macOS
7. Repeat the test using the Camera

***iOS connect from iOS***

1. Create an account on device A
2. Delete any existing account from device B
3. On Device B turn on sync and continue through to sync another device
4. Tap show QR Code
5. Copy the code
6. On Device A tap on Scan QR Code
7. Paste it in as in the macOS flow
8. After a few seconds the account should be connected
9. Repeat with Camera

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/DDGSync/internal/RecoveryKeyTransmitter.swift
+++ b/Sources/DDGSync/internal/RecoveryKeyTransmitter.swift
@@ -1,0 +1,61 @@
+//
+//  RecoveryKeyTransmitter.swift
+//  DuckDuckGo
+//
+//  Copyright © 2023 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct RecoveryKeyTransmitter: RecoveryKeyTransmitting {
+
+    let endpoints: Endpoints
+    let api: RemoteAPIRequestCreating
+    let storage: SecureStoring
+    let crypter: Crypting
+
+    func send(_ code: SyncCode.ConnectCode) async throws {
+        guard let account = try storage.account() else {
+            throw SyncError.accountNotFound
+        }
+
+        guard let token = try storage.account()?.token else {
+            throw SyncError.noToken
+        }
+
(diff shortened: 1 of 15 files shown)


Sources/DDGSync/internal/RemoteConnector.swift
```diff
@@ -0,0 +1,109 @@
...
+//
+
+import Foundation
+
+struct RemoteConnector: RemoteConnecting {
+    
+    let code: String
+    let connectInfo: ConnectInfo
+
+    let account: AccountManaging
```
ayoy, line 26: `account` and `storage` seem unneeded in this struct.

ayoy (approved):
Looks great! Only a couple of minor comments.

Sources/DDGSync/internal/RecoveryKeyTransmitter.swift
```diff
@@ -0,0 +1,63 @@
...
+        let body = try JSONEncoder.snakeCaseKeys.encode(
+            ConnectRequest(deviceId: code.deviceId, encryptedRecoveryKey: encryptedRecoveryKey)
+        )
+
+        let request = api.createRequest(url: endpoints.connect,
+                                        method: .POST,
+                                        headers: ["Authorization": "Bearer \(token)"],
+                                        parameters: [:],
+                                        body: body,
+                                        contentType: "application/json")
```
samsymons, line 54: Tiny nitpick, this feels to me like the user of the API code has to care too much about the format of the request (like having to format the authorization header manually, and specify the content type string).

I assume we'll be needing to make similar calls in a number of other places, so it would be nice to have it handled more succinctly from the caller's perspective. What do you think?

brindy: Agree the authorization header could be hidden away (other headers are already added elsewhere), but everything else is the responsibility of the api call to set, so laying it out like this shows the intent.  If I were to change anything anything it might be to do something like this:

`api.createAuthenticatedRequest(...)`

Maybe body could be more explicitly `jsonBody` then there'd be no need to set the content type either, I suppose.

I think we should see what we end up doing for the actual data sync and then refactor then.  I don't really want to define something now that might get changed later tbh.

ayoy (approved):
LGTM! Love to see polling mechanism in BSK as originally planned 👏

PR #317, opened 2023-04-13
Deadlock fix: Do not listen on main queue

bwaresiak:

**Required**:

Task/Issue URL: https://app.asana.com/0/414709148257752/1204381704707794/f
iOS PR: 
macOS PR: 

**Optional**:

Tech Design URL:
CC:

**Description**:

Fix potential deadlock situation.

**Steps to test this PR**:

Break in notification handler and validate the threads.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/Bookmarks/MenuBookmarksViewModel.swift
+++ b/Sources/Bookmarks/MenuBookmarksViewModel.swift
@@ -48,6 +48,8 @@ public class MenuBookmarksViewModel: MenuBookmarksInteracting {
         }
         return _favoritesFolder
     }
+
+    private var observer: NSObjectProtocol?
     
     private let errorEvents: EventMapping<BookmarksModelError>?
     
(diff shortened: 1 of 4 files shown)


Sources/Bookmarks/MenuBookmarksViewModel.swift
```diff
@@ -58,16 +60,25 @@ public class MenuBookmarksViewModel: MenuBookmarksInteracting {
...
+            NotificationCenter.default.removeObserver(observer)
+            self.observer = nil
+        }
+    }
+
     private func registerForChanges() {
-        NotificationCenter.default.addObserver(forName: NSManagedObjectContext.didSaveObjectsNotification,
-                                               object: nil,
-                                               queue: .main) { [weak self] notification in
+        observer = NotificationCenter.default.addObserver(forName: NSManagedObjectContext.didSaveObjectsNotification,
```
brindy, line 71: Why not use the selector version now?  And then you also don't need all this extra code for unregistering, right?

https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver

bwaresiak: I was thinking about that, and leaned towards not doing so:
- I don't like having to mark function as @objc just to use that API :)
- This shows good example of having to remove observer when using this queue based API - so I though about keeping that in the codebase, as it is easy to forget that some of the Notification Center API does handle unregistering on its own, and some not.

I don't have strong opinion tho, so if you care/feel it should be changed I can do this.

brindy: Nah, I don't feel that strongly either :)

Bunn (approved):
LGTM.
Tested bookmarks on both iOS and macOS, as well as trying to reproduce the issue by closing/opening the app multiple times.

PR #321, opened 2023-04-18
Avoid direct calls to entity() on a ManagedObject subclass

bwaresiak:

**Required**:

Task/Issue URL:  https://app.asana.com/0/414235014887631/1204410214610010/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1651
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1134

**Description**:

Avoid direct calls to entity() on a ManagedObject subclass

**Steps to test this PR**:

General smoke tests for related functionalities.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SmarterEncryption/Store/HTTPSStoredBloomFilterSpecification.swift
+++ b/Sources/BrowserServicesKit/SmarterEncryption/Store/HTTPSStoredBloomFilterSpecification.swift
@@ -27,6 +27,15 @@ public class HTTPSStoredBloomFilterSpecification: NSManagedObject {
         return NSFetchRequest<HTTPSStoredBloomFilterSpecification>(entityName: "HTTPSStoredBloomFilterSpecification")
     }
 
+    public class func entity(in context: NSManagedObjectContext) -> NSEntityDescription {
+        return NSEntityDescription.entity(forEntityName: "HTTPSStoredBloomFilterSpecification", in: context)!
+    }
+
+    public convenience init(context moc: NSManagedObjectContext) {
+        self.init(entity: Self.entity(in: moc),
+                  insertInto: moc)
+    }
+
     @NSManaged public var bitCount: Int64
     @NSManaged public var errorRate: Double
     @NSManaged public var sha256: String?
--- a/Sources/BrowserServicesKit/SmarterEncryption/Store/AppHTTPSUpgradeStore.swift
+++ b/Sources/BrowserServicesKit/SmarterEncryption/Store/AppHTTPSUpgradeStore.swift
@@ -153,7 +153,7 @@ public struct AppHTTPSUpgradeStore: HTTPSUpgradeStore {
         context.performAndWait {
             deleteBloomFilterSpecification()
 
-            let storedEntity: HTTPSStoredBloomFilterSpecification = context.insertObject()
+            let storedEntity = HTTPSStoredBloomFilterSpecification(context: context)
             storedEntity.bitCount = Int64(specification.bitCount)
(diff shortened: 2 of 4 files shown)


Sources/BrowserServicesKit/SmarterEncryption/Store/HTTPSStoredBloomFilterSpecification.swift
```diff
@@ -27,6 +27,15 @@ public class HTTPSStoredBloomFilterSpecification: NSManagedObject {
...
 
+    public class func entity(in context: NSManagedObjectContext) -> NSEntityDescription {
+        return NSEntityDescription.entity(forEntityName: "HTTPSStoredBloomFilterSpecification", in: context)!
+    }
+
+    public convenience init(context moc: NSManagedObjectContext) {
+        self.init(entity: Self.entity(in: moc),
+                  insertInto: moc)
+    }
+
```
Bunn, line 38: Instead of adding these functions to both classes, what do you think about doing a generic extension?

```swift
extension NSManagedObject {
    public class func entity(in context: NSManagedObjectContext) -> NSEntityDescription {
        return NSEntityDescription.entity(forEntityName: entity().name!, in: context)!
    }

    public convenience init(context moc: NSManagedObjectContext) {
        self.init(entity: Self.entity(in: moc),
                  insertInto: moc)
    }
}
```
We do something similar [here](https://github.com/duckduckgo/iOS/blob/2c57fcfca7fd57fb895869c96ae130339f8d034e/Core/NSManagedObjectContextExtension.swift#L28)

bwaresiak: Well, I'm not opposed to that (especially sharing the first method) but the second is a bit more tricky and that's why I was reluctant to do so:

Basically `init(context:)` is a convenience init on `NSManagedObject`, meaning that extension would override it - which is supposedly fine, but then it also needs to be declared with @objc so classes derived from `NSManagedObject` can override that init if needed - unless we don't want that to happen which is also a valid approach. :)

I'm happy to go with that change tho, if we can agree that is is the good solution. Also I'd prefer to make it so `init(context:)` cannot be overridden in our subclasses.

Bunn: Cool, I'm honestly fine with either way, since @samsymons is going to review this I think we can wait for his thoughts before agreeing with this change or not.

samsymons: Since it's such little duplication I'm fine with the current approach, as we're not boxing ourselves in by merging it as-is. I'll defer it to Bartek whether he feels energetic enough to make the change before vacation begins. 🙂

samsymons (approved):

Sources/BrowserServicesKit/SmarterEncryption/Store/HTTPSStoredBloomFilterSpecification.swift
bwaresiak: Actually, we can't really put that into BSK without triggering error about ambiguous usage - as extension and class definitions itself would be imported (or not) depending on imports and dependencies. 

I'll be keeping the duplication for now.

PR #322, opened 2023-04-19
Update Bookmarks with isPendingDeletion and lastModified

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1204422863132217/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1653
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1136

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1204345998879965/f
CC: @bwaresiak 

**Description**:
This change updates Bookmarks data model:
* adds isPendingDeletion boolean flag – set to true to soft delete the object
* adds lastModified timestamp – set automatically to current date on willSave

Bookmarks database cleaner class is added to remove soft-deleted bookmarks on schedule.

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/Bookmarks/BookmarkEntity.swift
```diff
@@ -143,6 +157,19 @@ public class BookmarkEntity: NSManagedObject {
...
+        var queue: [BookmarkEntity] = [self]
+
+        while !queue.isEmpty {
+            let currentObject = queue.removeFirst()
+            currentObject.isPendingDeletion = true
+
+            if currentObject.isFolder {
+                queue.append(contentsOf: currentObject.childrenArray)
+            }
+        }
```
bwaresiak, line 171: Based on the privacy discussion, we should be cleaning up URL a title right? I wonder where is a good place to do that, but since we are already having business logic here (tree-handling) it may be as well be here?

Sources/Bookmarks/BookmarksDatabaseCleaner.swift
```diff
@@ -0,0 +1,88 @@
...
+            let bookmarksPendingDeletion = BookmarkUtils.fetchBookmarksPendingDeletion(context)
+
+            for bookmark in bookmarksPendingDeletion {
+                context.delete(bookmark)
+            }
+
+            do {
+                try context.save()
+            } catch {
+                errorEvents?.fire(.init(coreDataError: error))
```
bwaresiak, line 72: No something to think about: we don't really have merge strategies set up, so we need to consider two things:
- Save failures here, caused by change coming from other context - unlikely as we are recreating the context but possible due to time window we have between fetch and save.
- Save failure in all other Bookmarks contexts - also should not happen often as we are merging but is possible.

I was initially thinking this will need to be brought up once we start saving sync context, but since we've introduced that async cleanup, we most likely shall be doing TD for that problem now.

bwaresiak (changes requested):
Nice work, left few comments inline. 👍

Sources/Bookmarks/BookmarkEntity.swift
ayoy: That’s right. Apologies that this code is half-baked. It doesn’t reflect the latest decisions like clearing URL and title. Let me fix it. Would you rather see tree handling in `BookmarkListViewModel.softDeleteBookmark`? The problem for me was that iOS and macOS use different APIs for bookmarks – iOS uses bookmark management code from BSK while macOS has its own BookmarkStore (with `remove(objectsWithUUIDs:)` method), so the common denominator is BookmarkEntity.

Still I believe that clearing URL and title is part of soft deletion, and it doesn’t make sense on its own without also setting `isPendingDeletion` flag – so it should be added here.

Sources/Bookmarks/BookmarkListViewModel.swift
```diff
@@ -122,6 +122,19 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
...
+        }
+
+        bookmark.markPendingDeletion()
+
+        save()
+
+        bookmarks = parentFolder.childrenArray
+    }
+
     public func deleteBookmark(_ bookmark: BookmarkEntity) {
```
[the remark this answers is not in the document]
ayoy: Yes, we should, but I wasn’t initially sure about it. I’m going to remove this method.

Sources/Bookmarks/BookmarkEntity.swift
```diff
@@ -68,6 +71,16 @@ public class BookmarkEntity: NSManagedObject {
         isFavorite = false
     }
 
+    public override func willSave() {
+        let changedKeys = changedValues().keys
+        guard !changedKeys.isEmpty, !changedKeys.contains(NSStringFromSelector(#selector(getter: modifiedAt))) else {
+            return
+        }
+        if modifiedAt == nil {
```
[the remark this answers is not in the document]
ayoy: Oh, good point :) I suppose the check for `modifiedAt == nil` should not be here. Happy to add tests to verify it.

Sources/Bookmarks/BookmarkEntity.swift
bwaresiak: Well, we had been clearing the folder contents with `cascade` delete rule before these changes, so let's keep it here, along with other soft-deletion logic.

The only problem I have, is that it binds three responsibilities here:
1. DB entity Deletion - this is expected.
2. Privacy constraints re stored objects (cleaning) - this is fine, part of the model conformation with high level sync mechanism.
3. Sync data validation constraints - this is something one may not see a direct connection with: we must ensure that the cleared out data still allows us to perform successful PATCH request, thus tying this with model validation on the backend. If endpoint were to expect non-optional title, that requirement would have to be respected here, which couples server validation with this place.

So my only worry is in the end #3 here, so I wonder if we should fall back to the model Spec in Asana and explicitly mention there that we tie any sync-model validation with validation that we have in the DB (for every field).

Sources/Bookmarks/BookmarksDatabaseCleaner.swift
ayoy: Following our Zoom call, I’ve added a mechanism that retries cleanup on context merge errors.

Sources/Bookmarks/BookmarkEntity.swift
ayoy: We’ve agreed on a Zoom call that handling #3 is about updating docs and keeping our apps’ model validations the same as sync model validations. Hence I’ve removed the Core Data validation for non-nil bookmark URL.

PR #331, opened 2023-04-25
fix disconnect device by id

brindy:


brindy:
Testing a local automation

PR #355, opened 2023-05-16
Sync Engine with support for syncing bookmarks

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1204606874917765/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1724
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1203

**Optional**:

Tech Design URL:
* https://app.asana.com/0/481882893211075/1204311618174648/f
* https://app.asana.com/0/481882893211075/1204345998879965/f
* https://app.asana.com/0/481882893211075/1204306755229293/f
* https://app.asana.com/0/481882893211075/1204448565097516/f
* https://app.asana.com/0/0/1204501868567932/f

**Description**:
Changes to bookmarks:
* Updated Bookmarks model to v3 with added `isPendingDeletion` and `modifiedAt`
* Bookmarks are now soft-deleted at all times. When sync is off, a database cleaner runs on schedule
  and removes soft-deleted bookmarks.
* Client UIs are updated to not display soft-deleted bookmarks

Changes to Sync:
* Replaced `isAuthenticated` with `SyncAuthState` enum (inactive, setupNewAccount, addNewDevice, active)
* Added `SyncScheduling` and `DataProviding` public protocols. Made `Crypting` public too.
* Added `SyncMetadata` Core Data model, `SyncMetadataStore` class and `SyncFeatureEntity` class representing a syncable model.
* Sync is triggered from clients by calling scheduler APIs. Scheduler decides internally when to start sync.
* SyncQueue is an actor and it can only run a single sync operation at a time.

New module – SyncDataProviders:
* For now only contains `BookmarkProvider`, it’s an implementation of `DataProviding` for handling bookmarks.
  The logic is fully reused in iOS and macOS apps.

**Recommended reading order**:
1. BookmarkEntity.swift, BookmarkUtils.swift, BookmarksDatabaseCleaner.swift
  (other changes under Bookmarks models are related to iOS app UI)
3. DDGSyncing.swift, DDGSync.swift, ProductionDependencies.swift
4. SyncScheduler.swift, SyncQueue.swift
5. BookmarksProvider.swift
6. The rest.

**Steps to test this PR**:
To speed up testing sync you may want to set `SyncScheduler.appLifecycleEventsDebounceInterval` to 1.

1. Run client apps from develop and import some bookmarks.
1. Switch to `dominik/sync-engine` branches on both clients and run the app again. Make sure that the app didn’t crash at startup due to failed migration.
1. Enable sync on one of the devices. Then enable sync on another device by adding it to the first device’s account.
1. Make sure that devices’ data was merged, i.e. added device contains both local and remote bookmarks and vice-versa.
1. Make changes to data and verify that they are synced according to the design.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

—
###### Internal references:


Sources/DDGSync/DDGSync.swift
```diff
@@ -161,31 +162,83 @@ public class DDGSync: DDGSyncing {
...
+                        } else {
+                            await syncQueue.setUpAndStartFirstSync()
+                        }
+                    }
+                }
+
+            syncDidFinishCancellable = syncQueue.syncDidFinishPublisher
+                .sink { [weak self] result in
+                    if case .success = result {
+                        self?.updateAuthState()
```
bwaresiak, line 216: Is this needed?

Sources/DDGSync/internal/SyncQueue.swift
```diff
@@ -0,0 +1,252 @@
...
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            for (feature, dataProvider) in dataProviders {
+                group.addTask { [weak self] in
+                    guard let self else {
+                        return
+                    }
+                    os_log(.debug, log: self.log, "Syncing %{public}s", feature.name)
+
+                    do {
+                        var result: SyncResult = try await self.makeResult(for: dataProvider, fetchOnly: fetchOnly)
```
bwaresiak, line 154: Can we split this into Request and Result, then encapsulate first within the latter?

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
```diff
@@ -0,0 +1,258 @@
...
+            var syncableBookmarks: [Syncable] = []
+            context.performAndWait {
+                let bookmarks = BookmarkUtils.fetchModifiedBookmarks(context)
+                syncableBookmarks = bookmarks.compactMap { try? Syncable(bookmark: $0, encryptedWith: crypter) }
+            }
+            continuation.resume(with: .success(syncableBookmarks))
+        }
+    }
+
+    public func handleInitialSyncResponse(received: [Syncable], clientTimestamp: Date, serverTimestamp: String?, crypter: Crypting) async throws {
```
bwaresiak, line 95: We have two places we branch to handle initial sync and deduplication: here and in entity parsing, I wonder if we can make it more unified (e.g. branch only in one place)

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
```diff
@@ -0,0 +1,258 @@
...
+        let insertedObjects = Array(context.insertedObjects).compactMap { $0 as? BookmarkEntity }
+        let updatedObjects = Array(context.updatedObjects.subtracting(context.deletedObjects)).compactMap { $0 as? BookmarkEntity }
+
+        try context.save()
+        (insertedObjects + updatedObjects).forEach { bookmarkEntity in
+            if let uuid = bookmarkEntity.uuid, !excludedUUIDs.contains(uuid) {
+                bookmarkEntity.modifiedAt = nil
+            }
+        }
+        try context.save()
```
bwaresiak, line 243: We rather want to avoid double-saving as it could lead to weird issues in case latter save fails.

Let's do what we talked about: set modifiedAt to nil only for known (and up-front specified) set of objects based on a union of sent and received ones. Then filter out these with modifiedAt newer than sync stat date, and just clean that timestamp for the ones that are left.

As for how to mark these object, I think it is acceptable to have in-memory flag associated with manage object entity that is set before save and reset afterwards.

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
ayoy: Disclaimer that we need to stick to 2 public functions, separate for handling initial sync and regular sync. That’s because empty sent array doesn’t mean it’s an initial sync.

We can, however, unify implementations of `handleSyncResponse` and `handleInitialSyncResponse` into a single function handling both scenarios (the only difference between the two is the value of `deduplicateEntities` flag that’s passed to `BookmarksResponseHandler`) which is what I’m going to do.

Sources/DDGSync/DDGSync.swift
ayoy: I figured out why it was there. This is a leftover from when I tried to have a single `startSync` method to handle both initial and regular sync. It was a bad decision and I’ve since switched to 2 methods (`startSync` and `setUpAndStartFirstSync`) but have never deleted this subscription. It’s now gone though :)

Sources/Bookmarks/BookmarksDatabaseCleaner.swift
```diff
@@ -0,0 +1,113 @@
...
+        triggerSubject.send()
+    }
+
+    func removeBookmarksPendingDeletion() {
+        let context = database.makeContext(concurrencyType: .privateQueueConcurrencyType)
+
+        context.performAndWait {
+            var saveError: Error?
+
+            while true {
```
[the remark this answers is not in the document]
ayoy: I added a limit of 5 save attempts (maybe too many? maybe 3 is enough? that’s perhaps irrelevant anyway :)) and am just propagating the merge error to clients. I’ve added a generic `syncBookmarksFailed` debug Pixel in clients but I consider it temporary until we define pixels.

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
ayoy: I’ll try adding more unit tests to cover various scenarios.

Sources/DDGSync/SyncMetadataStore.swift
```diff
@@ -0,0 +1,68 @@
...
+            lastModified = feature?.lastModified
+        }
+        return lastModified
+    }
+
+    public func updateTimestamp(_ timestamp: String?, forFeatureNamed name: String) {
+        context.performAndWait {
+            SyncFeatureUtils.updateTimestamp(timestamp, forFeatureNamed: name, in: context)
+
+            try? context.save()
```
bwaresiak, line 63: Can we also fire Pixel on errors from this store?

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
```diff
@@ -0,0 +1,232 @@
...
+
+            do {
+                try context.save()
+            } catch {
+                saveError = error
+            }
+        }
+
+        if let saveError {
+            throw saveError
```
bwaresiak, line 73: Aren't these errors ignored in the call chain?

ayoy: Those errors are propagated all the way up to `SyncQueue.handleResponse` and they’re being caught in [this catch clause](https://github.com/duckduckgo/BrowserServicesKit/blob/dominik/sync-engine/Sources/DDGSync/internal/SyncQueue.swift#L172-L176). That’s how I understand it :)

Sources/DDGSync/SyncMetadataStore.swift
ayoy: Yes, I’d add a separate pixel for that, what do you think?

ayoy: I added `m_d_sync_bookmarks_provider_initialization_failed`.

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
bwaresiak: I'm not sure that's the case - isn't this called from [here](https://github.com/duckduckgo/BrowserServicesKit/pull/355/files/c9dd8d2dc39833f828be3c7ebbf4a4b6baba9262#diff-7fead194fe4fc0c9151d18a137ec673550e65ddf3b22628bcaa1e5b107653020R200)? And not all of the callse to update account are handled.

ayoy: My bad, I confused this code with a similar one in `handleSyncResponse`. You’re right, I’ll tackle this tomorrow.

ayoy: Ok, I handled it by catching this error in SyncQueue and notifying dataProvider error handler (it ends up being handled the same way as errors thrown during sync). Let me know if you’d prefer a different approach.

bwaresiak: This is good :)

Sources/Bookmarks/BookmarksDatabaseCleaner.swift
```diff
@@ -0,0 +1,120 @@
...
+        let context = database.makeContext(concurrencyType: .privateQueueConcurrencyType)
+        var saveAttemptsLeft = Const.maxContextSaveRetries
+
+        context.performAndWait {
+            var saveError: Error?
+
+            while true {
+                let bookmarksPendingDeletion = fetchBookmarksPendingDeletion(context)
+                if bookmarksPendingDeletion.isEmpty {
+                    print("No bookmarks pending deletion")
```
bwaresiak, line 74: Shouldn't these be os logs? :)

ayoy: Sure, I forgot to update them, let me handle this now :)

bwaresiak (approved):
Great job!

PR #372, opened 2023-05-26
Autofill hash migration updates + Bitwarden hotfix

afterxleep:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1177771139624306/1204667757121868/f && https://app.asana.com/0/1204099484721401/1204677118392502/f

iOS PR: 
macOS PR:  https://github.com/duckduckgo/macos-browser/pull/1226

**Description**:
- Optimizes account hashing migration
- Stop removing account duplicates for autofill when the signature is nil

**Steps to test this PR**:
1.  See macOS PR for testing instructions

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/SecureVaultCryptoProvider.swift
+++ b/Sources/BrowserServicesKit/SecureVault/SecureVaultCryptoProvider.swift
@@ -32,9 +32,20 @@ protocol SecureVaultCryptoProvider {
     func encrypt(_ data: Data, withKey key: Data) throws -> Data
 
     func decrypt(_ data: Data, withKey key: Data) throws -> Data
-        
+
     func hashData(_ data: Data) throws -> String?
 
+    func hashData(_ data: Data, salt: Data?) throws -> String?
+
+    var hashingSalt: Data? { get }
+
+}
+
+extension SecureVaultCryptoProvider {
+    func hashData(_ data: Data) throws -> String? {
+        guard let salt = hashingSalt else { throw SecureVaultError.failedToGetHashingSalt }
+        return try hashData(data, salt: salt)
+    }
 }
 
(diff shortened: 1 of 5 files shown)


Sources/BrowserServicesKit/SecureVault/SecureVaultCryptoProvider.swift
```diff
@@ -170,14 +181,14 @@ final class DefaultCryptoProvider: SecureVaultCryptoProvider {
...
     }
 
-        
-    func hashData(_ data: Data) throws -> String? {
-        if let salt = self.salt {
-            let saltedData = salt + data
-            return SHA256.hash(data: saltedData).dataRepresentation.base64EncodedString()
+    func hashData(_ data: Data, salt: Data? = nil) throws -> String? {
+        guard let salt = salt ?? hashingSalt else {
+            throw SecureVaultError.failedToGetHashingSalt
```
bwaresiak, line 186: In previous implementation, salt could be nil and then this function would return nil, in the new implementation it throws an error. This will potentially have effect on the methods like `storeWebsiteCredentials` (breaking the flow instead of silently storing nil signature). I think it is a good change (as it falls under error handling we have there), just want to double check if you foresee any potential issues with that.

bwaresiak (approved):
Migration is much faster now, LGTM!

Sources/BrowserServicesKit/SecureVault/SecureVaultCryptoProvider.swift
afterxleep: @bwaresiak   I don't see any issues, but from a peace of mind (and since this is a hotfix), I can revert this change for release, and add it back after confirming everything is behaving correctly in the wild.  WDYT?

PR #373, opened 2023-05-30
Ignore bookmarks with missing UUIDs in sync

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1204699272468090/f

**Description**:
When UUID is missing, throw an error in Syncable initializer. The error is silenced up the call chain
and the broken bookmark ends up not being included in sync.

**Steps to test this PR**:
It's tricky to simulate a crash programmatically now because you can't save a context when bookmark is missing a UUID.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/duckduckgo/content-scope-scripts",
       "state" : {
-        "revision" : "801b7f23476f797c6eaa72b070e6c80abb82801a",
-        "version" : "4.4.4"
+        "revision" : "260329ad8021115235d64ddd76c24629f93a7212",
+        "version" : "4.21.1"
       }
     },
     {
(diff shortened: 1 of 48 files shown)


samsymons:
> What kind of version bump will this require?: Patch

Pedantic question about semantic versioning: since this PR introduces new public types, should it instead be a minor version bump?

ayoy:
Good catch @samsymons! In this case we're not adding any public symbols so patch bump makes sense to me. Also the added error is not really propagated to clients. The removed API is also internal, so it doesn't affect public API.

ayoy:
> > so let me make it internal :)
> 
> Which means that this now needs a major version bump since that's technically breaking API change 🙃

I'm not making any existing public APIs internal, only adding a new internal type and removing 2 internal functions, so I think joke's on you after all ;)

samsymons (approved):
LGTM, I set up sync with this branch of BSK in both iOS and macOS and generally did some smoke testing, no issues found.

I did run into the case below in the macOS browser, but I don't think it's relevant to this change as I just ran this branch against macOS develop.

CleanShot 2023-06-26 at 21 59 54@2x

ayoy:
@samsymons thanks for the review and apologies for not creating platform branches! 🙈 Can confirm that the assertion being thrown on macOS is not relevant to this change as I've also stumbled upon it last week on develop.

PR #383, opened 2023-06-15
Reattach orphaned bookmarks to root folder when reordering

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1204720539208728/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1771
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1260

**Description**:
When moving a bookmark inside root folder, consider orphaned bookmarks when validating toIndex and set
parent to all affected orphaned bookmarks. See unit tests for detailed logic.

This change extracts BookmarkTree into a separate Swift module called BookmarksTestsUtils, so that it can
be used for testing Bookmarks module.

**Steps to test this PR**:
1. Verify that unit test pass

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/Bookmarks/BookmarkListViewModel.swift
+++ b/Sources/Bookmarks/BookmarkListViewModel.swift
@@ -250,6 +250,9 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
 
         if shouldFetchRootFolder {
             let orphanedBookmarks = BookmarkUtils.fetchOrphanedEntities(context)
+            if !orphanedBookmarks.isEmpty {
+                errorEvents?.fire(.orphanedBookmarksPresent)
+            }
             folderBookmarks += orphanedBookmarks
         }
         return folderBookmarks
--- a/Sources/Bookmarks/BookmarkListViewModel.swift
+++ b/Sources/Bookmarks/BookmarkListViewModel.swift
@@ -145,7 +145,7 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
             return
         }
 
-        let orphanedBookmarks: [BookmarkEntity] = BookmarkUtils.fetchOrphanedEntities(context)
+        let orphanedBookmarks = bookmarks.filter { $0.parent == nil }
         guard !orphanedBookmarks.isEmpty else {
             return
         }
--- a/Sources/Bookmarks/BookmarkListViewModel.swift
+++ b/Sources/Bookmarks/BookmarkListViewModel.swift
@@ -107,12 +107,42 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
         save()
     }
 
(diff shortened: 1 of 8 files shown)


Sources/Bookmarks/BookmarkListViewModel.swift
```diff
@@ -138,6 +140,30 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
         refresh()
     }
 
+    private func reattachOrphanedBookmarks(forMoving bookmark: BookmarkEntity, toIndex: Int) {
+        guard let rootFolder = BookmarkUtils.fetchRootFolder(context) else {
+            return
+        }
+
+        let orphanedBookmarks: [BookmarkEntity] = BookmarkUtils.fetchOrphanedEntities(context)
```
bwaresiak, line 148: Do we need to fetch this? Can't we just filter displayed bookmarks to get these that have no parent? I'm afraid if refetching could still lead to index inconsistencies if we have changes happening in the background.

ayoy: Of course we can! Thanks for spotting this, fixing now :)

ayoy: It's fixed and I've also added a pixel to track orphaned bookmarks presence for users not on sync. It's ready for review 🙏

Sources/Bookmarks/BookmarkListViewModel.swift
```diff
@@ -138,6 +140,30 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
...
+            return
+        }
+
+        let orphanedBookmarksToAttachToRootFolder: [BookmarkEntity] = {
+            let toIndexInOrphanedBookmarks = toIndex - rootFolder.childrenArray.count
+            guard bookmark.parent == nil else {
+                return toIndexInOrphanedBookmarks >= 0 ? Array(orphanedBookmarks.prefix(through: toIndexInOrphanedBookmarks)) : []
+            }
+            guard let bookmarkIndexInOrphans = orphanedBookmarks.firstIndex(where: { $0.uuid == bookmark.uuid }) else {
+                return [bookmark]
```
bwaresiak, line 159: If the bookmarks is not in orphans, we do ned to reattach it?

bwaresiak: Ah I see but we fallen through the parent check so the bookmark itself is orphan, this guard should never happen.

ayoy: Exactly, especially now that we're not refetching orphans, but just taking all without a parent. Can as well return an empty array here (and perhaps an assertion failure), but the expectation is that this doesn't happen.

bwaresiak (approved):
LGTM!

PR #396, opened 2023-06-30
Manage Private Email Addresses

afterxleep:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1204615490114675/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1828
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1319

**Optional**:

**Description**:
Allows users to manage their Private Email addresses from the browser.

**Steps to test this PR**:
1. Check macOS/iOS PRs and test functionality
2. Check iOS PR and smoke test password generation and credentials saving

---
###### Internal references:


Sources/BrowserServicesKit/Autofill/AutofillUserScript+SecureVault.swift
```diff
@@ -484,19 +480,6 @@ extension AutofillUserScript {
 
         let domain = hostForMessage(message)
 
-        if request.mainType == .credentials, request.subType == .password, let generatedPassword = request.generatedPassword?.value, !generatedPassword.isEmpty {
```
amddg44, line 487: @afterxleep this code is needed for password generation on iOS (to notify the native layer to display the autofill prompt and return back the users selected action)

afterxleep: you're right.. added this method back @amddg44

Sources/BrowserServicesKit/SecureVault/SecureVaultManager.swift
```diff
@@ -53,11 +52,6 @@ public protocol SecureVaultManagerDelegate: SecureVaultErrorReporting {
                             promptUserWithGeneratedPassword password: String,
                             completionHandler: @escaping (Bool) -> Void)
 
-    func secureVaultManagerShouldAutomaticallyUpdateCredentialsWithoutUsername(_: SecureVaultManager,
```
amddg44, line 56: I still see one or both of these delegate methods implemented in the codebases on iOS / macOS and in MockSecureVaultManagerDelegate; it would good to get these cleaned up 🙂

amddg44 (changes requested):
@afterxleep I've tested both platforms thoroughly and it's all looking really good + loving the cleanup of the logic around auto-saving / updating! Just a couple of small comments 👍

afterxleep:
@amddg44 Requested changes done, thank you for the feedback

afterxleep:
@amddg44 

I have pushed the following changes:

1. If a form had some autogenerated data at some point it should remain that way until submission.  I added a check to change the `autogeneratedCredentials`  only if it's false, [here](https://github.com/duckduckgo/BrowserServicesKit/pull/396/commits/0b92c9393ec84c206e0bbd6e970fd949ac2913a9#diff-2683a453a2ed1a0cd025ce1941481dab5164acb09ab4f5e1c1e6e58d04ffa2baR189)
2. If the user deleted an account what was autoSaved, some logic may stop working, and this could cause issues, like the one you experienced when the username was not saved properly, so I'm now checking for the validity of that account [here](https://github.com/duckduckgo/BrowserServicesKit/pull/396/commits/0b92c9393ec84c206e0bbd6e970fd949ac2913a9#diff-2683a453a2ed1a0cd025ce1941481dab5164acb09ab4f5e1c1e6e58d04ffa2baR179)
3. I removed the remaining [autodelete/autosave logic from the iOS TabViewController](https://github.com/duckduckgo/iOS/pull/1828/files#diff-23f45a546b44732f442ecaff49aa3d7bbdb120eadd9cf312fc61b8d5181b6ad0L2419) .  Partial account deletion is handled by BSK, [here](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-2683a453a2ed1a0cd025ce1941481dab5164acb09ab4f5e1c1e6e58d04ffa2baR215).
4., I updated the submission logic to perform some comparisons between partial logins and submitted data, which allows us to find out if the user has manually changed autogenerated usernames or passwords.  This was added [here](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-2683a453a2ed1a0cd025ce1941481dab5164acb09ab4f5e1c1e6e58d04ffa2baR194)

5. I added a good set of tests to cover to the top scenarios around autosaving emails and passwords, and removed old ones that were incomplete (not accounting for emails, for example. (Incluiding scenario 1).  You can find them [here](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-c42229b6ba0ddb79302af9009d06c791f7d56ddf3c90bf43c86a7be3d81fd938R236)

afterxleep:
@amddg44  I also added a good set of tests that should cover all autosaving scenarios for emails and passwords, and removed old ones that were incomplete (not accounting for emails, for example).   You can find them [here](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-c42229b6ba0ddb79302af9009d06c791f7d56ddf3c90bf43c86a7be3d81fd938R236)

afterxleep:
@amddg44 

A few changes to fix the reported issues:

1. [Usernames are saved when generating a password (if present) + test](https://github.com/duckduckgo/BrowserServicesKit/pull/396/commits/5f5af9c64aab308d7754bb7ac70a781b32b4820b)

2. When creating a private email address and then switching to a manual address, a prompt was presented, but you ended up with two accounts.  I added [a test](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-c42229b6ba0ddb79302af9009d06c791f7d56ddf3c90bf43c86a7be3d81fd938R418) for this which passes in the MockVaultProvider, but apparently the DefaultSecureVault fails.   [Moving temporary account deletion to the main function](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-2683a453a2ed1a0cd025ce1941481dab5164acb09ab4f5e1c1e6e58d04ffa2baR224) fixes the issue.  This was likely caused by an unexpected vault dealloc. ??

3. When generating a password on a site and navigating away to another login form in a different domain credentials were autosaved, even if no data was autogenerated.   I have added [logic to clear up autosaveAccount and autogenerated username and pass bools when navigating away](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-2683a453a2ed1a0cd025ce1941481dab5164acb09ab4f5e1c1e6e58d04ffa2baR185), and added [a test](https://github.com/duckduckgo/BrowserServicesKit/pull/396/files#diff-c42229b6ba0ddb79302af9009d06c791f7d56ddf3c90bf43c86a7be3d81fd938R557) for coverage.

afterxleep:
@amddg44 

As for the reported issue "3f": (Existing account being auto-updated when generating password), the behavior on Android is to create another account.  

While we cannot do that, due to multiple account with the same username constraint, I'm no longer autosaving data for those accounts.   This should trigger an update credentials dialog when a new password is added.

Screen Recording 2023-08-01 at 8 26 17 PM

afterxleep:
@amddg44 

As for the reported issue "3f": (Existing account being auto-updated when generating password), the behavior on Android is to create another account.

While we cannot do that due to the unique username constraint, I'm no longer auto-updating data for those accounts, which should trigger a update dialog on submission.   This is also covered by a unit test

Screen Recording 2023-08-01 at 8 27 28 PM

Sources/BrowserServicesKit/Email/EmailManager.swift
```diff
@@ -242,6 +260,9 @@ public class EmailManager {
         return username + "@" + EmailManager.emailDomain
     }
 
+    // Whethe we should autosave email addresses in the vault
+    private var shouldAutosave: Bool = true
```
amddg44, line 264: I don't think this variable is used anywhere

Sources/BrowserServicesKit/SecureVault/SecureVaultManager.swift
```diff
@@ -103,6 +97,20 @@ public class SecureVaultManager {
 
     public let tld: TLD?
 
+    // Keeps track of partial account created from autogenerated credentials (Private Email + Pwd)
+    public var autosaveAccount: SecureVaultModels.WebsiteAccount?
```
amddg44, line 101: Could all of these new vars be changed to `private`?

Sources/BrowserServicesKit/SecureVault/SecureVaultManager.swift
```diff
@@ -400,93 +476,77 @@ extension SecureVaultManager: AutofillSecureVaultDelegate {
...
+            currentAccount = createAccount(vault: vault, username: user,
+                                                 password: Data((pass).utf8),
+                                                 domain: domain)
+            autosaveAccountCreatedInSession = true
         }
-        
-        if accounts.contains(where: { $0.username == autofillCredentials.username }) {
-            os_log("ERROR: Tried to save generated credentials with an existing username, prompt the user to update instead.", log: .passwordManager)
+
+        // Automatically update user/pass id the current accout was created in this session
```
amddg44, line 510: Tiny nit, can this comment get updated? It's a little hard to understand 🙂

amddg44 (approved):
LGTM! This is a significant improvement to the logic around autosaving; great work @afterxleep 🎉

PR #408, opened 2023-07-07
Fix bookmarks reordering

bwaresiak:

**Required**:

Task/Issue URL: https://app.asana.com/0/414235014887631/1204953708712054/f
iOS PR: 
macOS PR: 

**Description**:

Fix reordering of bookmarks in case there are objects pending deletion.

**Steps to test this PR**:

Delete first & last bookmarks from the list, then start reordering other ones.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/Bookmarks/BookmarkListViewModel.swift
+++ b/Sources/Bookmarks/BookmarkListViewModel.swift
@@ -139,6 +139,12 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
         let actualFromIndex = mutableChildrenSet.index(of: bookmark)
         let actualToIndex = mutableChildrenSet.index(of: visibleChildren[toIndex])
 
+        guard actualFromIndex != NSNotFound, actualToIndex != NSNotFound else {
+            assertionFailure("Bookmark: position could not be determined")
+            refresh()
+            return
+        }
+
         mutableChildrenSet.moveObjects(at: IndexSet(integer: actualFromIndex), to: actualToIndex)
 
         save()
--- a/Sources/Bookmarks/BookmarkListViewModel.swift
+++ b/Sources/Bookmarks/BookmarkListViewModel.swift
@@ -120,21 +120,26 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
             return
         }
 
-        guard let children = parentFolder.children,
-              fromIndex < children.count,
-              toIndex < children.count else {
+        let visibleChildren = parentFolder.childrenArray
+
+        guard fromIndex < visibleChildren.count,
+              toIndex < visibleChildren.count else {
             errorEvents?.fire(.indexOutOfRange(.bookmarks))
             return
         }
 
-        guard let actualBookmark = children[fromIndex] as? BookmarkEntity,
-              actualBookmark == bookmark else {
+        guard visibleChildren[fromIndex] == bookmark else {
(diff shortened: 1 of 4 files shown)


Sources/Bookmarks/BookmarkListViewModel.swift
```diff
@@ -120,21 +120,26 @@ public class BookmarkListViewModel: BookmarkListInteracting, ObservableObject {
...
             errorEvents?.fire(.bookmarksListIndexNotMatchingBookmark)
             return
         }
 
+        // Take into account bookmarks that are pending deletion
         let mutableChildrenSet = parentFolder.mutableOrderedSetValue(forKeyPath: #keyPath(BookmarkEntity.children))
-        mutableChildrenSet.moveObjects(at: IndexSet(integer: fromIndex), to: toIndex)
+
+        let actualFromIndex = mutableChildrenSet.index(of: bookmark)
+        let actualToIndex = mutableChildrenSet.index(of: visibleChildren[toIndex])
```
ayoy, line 140: These calls may return `NSNotFound` (which I believe equals -1), but `IndexSet` initializer fails when initialized with a negative index. I realize it's unlikely to happen, but perhaps it's still worth to guard against it?

Tests/BookmarksTests/BookmarkListViewModelTests.swift
```diff
@@ -77,6 +77,45 @@ final class BookmarkListViewModelTests: XCTestCase {
...
+        }
+
+        context.performAndWait {
+            bookmarkTree.createEntities(in: context)
+
+            let deleted1 = BookmarkEntity.fetchBookmark(withUUID: "1", context: context)!
+            deleted1.markPendingDeletion()
+
+            let deleted3 = BookmarkEntity.fetchBookmark(withUUID: "3", context: context)!
+            deleted3.markPendingDeletion()
```
ayoy, line 98: The following is equivalent (you don't need to fetch individual bookmarks and mark them as deleted):

```suggestion
        let bookmarkTree = BookmarkTree {
            Bookmark(id: "1", isDeleted: true)
            Bookmark(id: "2")
            Bookmark(id: "3", isDeleted: true)
            Bookmark(id: "4")
        }

        context.performAndWait {
            bookmarkTree.createEntities(in: context)
```

ayoy (approved):
LGTM, works great! I tested it on and iOS device with disabled sync.

Leaving a comment on error handling, in case you missed it, and another one about simplifying tests code.

Sources/Bookmarks/BookmarkListViewModel.swift
bwaresiak: As `visibleChildren` call above already fetched `children` relationship, this one should not return different (smaller) result set, and both bounds and existence if bookmark has been validated earlier. I'd be surprised to see this fail and I wanted to avoid having to guard and report another pixel here.

ayoy: Sure, perhaps an assertion makes sense to make intentions clear?

bwaresiak: I've added assertion, and applied same patter to favorites.

ayoy: Awesome, thanks!

PR #435, opened 2023-07-27
Support Config v3 and incremental feature flags

SlayterDev:

**Required**:

Task/Issue URL: https://app.asana.com/0/1201470200663180/1205012648854940/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1870
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1400

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1205076737635363/f
CC:

**Description**:
This PR adds support for config v3 by adding incremental feature flags. There are no features to test with yet but the mechanism is unit tested here.

**Steps to test this PR**:
1. Verify unit tests
1.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -88,14 +96,71 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+                                rollouts: [PrivacyConfigurationData.PrivacyFeature.Feature.Rollout],
+                                randomizer: (Range<Double>) -> Double) -> Bool {
+        let defsPrefix = "config.\(subfeature.parent.rawValue).\(subfeature.rawValue)"
+        if userDefaults.bool(forKey: "\(defsPrefix).\(Constants.enabledKey)") {
+            return true
+        }
+        
+        var willEnable = false
+        if let rolloutSize = userDefaults.value(forKey: "\(defsPrefix).\(Constants.lastRolloutCountKey)") as? Int {
+            guard rolloutSize != rollouts.count else { return false }
```
samsymons, line 110: I wonder here whether we should check that the new rollout size is larger than the old one, rather than just different.

The situation I have in mind (and it's an edge case for sure) is that we deploy a new rollout value - but quickly remove it. With the current implementation, that means that any clients that got a new config in the window where the new rollout value was available will update again and have yet another chance to roll.

By making this `newRolloutSize > oldRolloutSize`, then we can at least quickly remove a rollout value to prevent new clients from getting the new rollout value, but any clients that managed to get it won't re-roll.

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -88,14 +96,71 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+            // First time user sees feature, and multiple rollouts
+            for i in 1..<rollouts.count {
+                let y = rollouts[i].percent
+                let x = rollouts[i - 1].percent
+                let prob = (y - x) / (100.0 - y)
+                if randomizer(0..<1) < prob {
+                    willEnable = true
+                    break
+                }
+            }
```
samsymons, line 136: I think we can just take the last step here and roll using that, right? Given steps `[5, 10, 15]`, we should just be able to take `15` and treat it as our threshold.

Sources/BrowserServicesKit/PrivacyConfig/PrivacyConfigurationData.swift
```diff
@@ -112,15 +112,37 @@ public struct PrivacyConfigurationData {
...
+            
+            public struct Rollout {
+                public let percent: Double
+                
+                enum CodingKeys: String {
+                    case percent
+                }
+                
+                public init(json: [String: Any]) {
+                    self.percent = json[CodingKeys.percent.rawValue] as? Double ?? 0
```
samsymons, line 126: Makes sense to fall back to `0` here :+1: I was wondering what would happen if there was an error parsing the rollout in the config (like if someone used `"5.0"` for the rollout value), and this way we won't make the feature visible unintentionally.

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -88,14 +96,71 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+        if let rolloutSize = userDefaults.value(forKey: "\(defsPrefix).\(Constants.lastRolloutCountKey)") as? Int {
+            guard rolloutSize != rollouts.count else { return false }
+            // Sanity check as we need at least two values to compute the new probability
+            guard rollouts.count > 1 else { return false }
+            
+            // If the user has seen the rollout before, and the rollout count has changed
+            // Try again with the new probability
+            let y = rollouts[rollouts.count - 1].percent
+            let x = rollouts[rollouts.count - 2].percent
+            let prob = (y - x) / (100.0 - y)
```
samsymons, line 118: If we somehow have a case where y is `100` this would lead to a divide by 0. That's real edge case territory though 🙂

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
SlayterDev: The spec says to compute it in this manner: https://app.asana.com/0/1203137811378537/1204934761241565/f

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
SlayterDev: I misread the spec here. We're supposed to subtract x from 100 here.

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
samsymons: That's true, though it's outdated - the Android implementation is doing it the simpler way, and Bartek has a comment [here](https://app.asana.com/0/0/1204934761241565/1204955479205817/f) suggesting this simpler approach also.

SlayterDev: Gotcha. Missed that. Updated now.

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -88,14 +96,64 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
         default: return false
         }
     }
+    
+    private func rolloutEnabled(subfeature: any PrivacySubfeature,
```
bwaresiak, line 100: Maybe `isRolloutEnabled` ?

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -88,14 +96,71 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+    public func isSubfeatureEnabled(_ subfeature: any PrivacySubfeature, versionProvider: AppVersionProvider, randomizer: (Range<Double>) -> Double) -> Bool {
         guard isEnabled(featureKey: subfeature.parent, versionProvider: versionProvider) else {
             return false
         }
         let subfeatures = subfeatures(for: subfeature.parent)
         let subfeatureData = subfeatures[subfeature.rawValue]
         let satisfiesMinVersion = satisfiesMinVersion(subfeatureData?.minSupportedVersion, versionProvider: versionProvider)
+        
+        // Handle Rollouts
+        if let rollouts = subfeatureData?.rollouts, !rollouts.isEmpty {
```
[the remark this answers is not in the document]
bwaresiak: I think check for length of the rollouts should be encapsulated in `isRolloutEnabled` as it seems part of the logic. That should be one of the border conditions we test.

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -88,14 +96,65 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+        // Empty rollouts should be default enabled
+        guard !rollouts.isEmpty else { return true }
+        
+        let defsPrefix = "config.\(subfeature.parent.rawValue).\(subfeature.rawValue)"
+        if userDefaults.bool(forKey: "\(defsPrefix).\(Constants.enabledKey)") {
+            return true
+        }
+        
+        var willEnable = false
+        let rollouts = rollouts.sorted(by: { $0.percent < $1.percent })
```
bwaresiak, line 112: Maybe just in case, we should be removing duplicates and filtering out values outside of 0...100 range?

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -21,21 +21,29 @@ import Foundation
...
     private let locallyUnprotected: DomainsProtectionStore
     private let internalUserDecider: InternalUserDecider
+    private let userDefaults: UserDefaults
 
     public init(data: PrivacyConfigurationData,
                 identifier: String,
                 localProtection: DomainsProtectionStore,
-                internalUserDecider: InternalUserDecider) {
+                internalUserDecider: InternalUserDecider,
+                userDefaults: UserDefaults = UserDefaults()) {
```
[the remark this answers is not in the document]
bwaresiak: I don't want to hold this PR for that - please create a follow up task to discuss that to make sure we are ok with this approach.

bwaresiak (approved):
LGTM!

PR #445, opened 2023-08-04
Create observers to extract from StatusReporter

graeme:

**Required**:

Task/Issue URL: 
iOS PR: 
macOS PR: 

**Optional**:

Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/Status/ConnectivityIssueObserver/ConnectivityIssueObserverThroughDistributedNotifications.swift
+++ b/Sources/NetworkProtection/Status/ConnectivityIssueObserver/ConnectivityIssueObserverThroughDistributedNotifications.swift
@@ -1,0 +1,84 @@
+//
+//  ConnectivityIssueObserverThroughDistributedNotifications.swift
+//
+//  Copyright © 2023 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+#if os(macOS)
+
+import Combine
+import Foundation
+import NetworkExtension
+import NotificationCenter
+import Common
+
+/// Observes the tunnel status through Distributed Notifications.
+///
+public class ConnectivityIssueObserverThroughDistributedNotifications: ConnectivityIssueObserver {
+    public lazy var publisher: AnyPublisher<Bool, Never> = subject.eraseToAnyPublisher()
+    public var recentValue: Bool {
+        subject.value
+    }
+
+    private let subject = CurrentValueSubject<Bool, Never>(false)
+
+    // MARK: - Notifications
+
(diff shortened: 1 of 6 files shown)


Sources/NetworkProtection/Status/ConnectivityIssueObserver/ConnectivityIssueObserverThroughDistributedNotifications.swift
```diff
@@ -0,0 +1,84 @@
...
+
+import Combine
+import Foundation
+import NetworkExtension
+import NotificationCenter
+import Common
+
+/// Observes the tunnel status through Distributed Notifications.
+///
+public class ConnectivityIssueObserverThroughDistributedNotifications: ConnectivityIssueObserver {
```
graeme, line 29: This is extracted from the `NetworkProtectionStatusReporter` in PR #446

Sources/NetworkProtection/Status/ControlllerErrorMessageObserver/ControllerErrorMesssageObserverThroughDistributedNotifications.swift
```diff
@@ -0,0 +1,85 @@
...
+
+import Combine
+import Foundation
+import NetworkExtension
+import NotificationCenter
+import Common
+
+/// Observes the tunnel status through Distributed Notifications.
+///
+public class ControllerErrorMesssageObserverThroughDistributedNotifications: ControllerErrorMesssageObserver {
```
graeme, line 29: This is extracted from the `NetworkProtectionStatusReporter` in #446

Sources/NetworkProtection/Status/ControlllerErrorMessageObserver/ControllerErrorMesssageObserver.swift
```diff
@@ -0,0 +1,26 @@
...
+//  limitations under the License.
+//
+
+import Combine
+import Foundation
+import NetworkExtension
+
+public protocol ControllerErrorMesssageObserver {
+    var publisher: AnyPublisher<String?, Never> { get }
+    var recentValue: String? { get }
```
samsymons, line 25: The naming here is a little vague, maybe something like `lastConnectionStatus`? Could go either way here tbh

graeme: I’m in two minds about this. I guess if we were to apply that thinking fully, the `publisher` should also be `controllerErrorPublisher`. But I’m not sure we’re going to get a lot out of these longer names as it’s pretty obvious what this is publishing + what type `recentValue` is from the name of this type.

Furthermore, I’m not sure I'm 100% happy with this refactor as I think we could remove a lot of the boilerplate with some more generic / polymorphism usage. But I wanted to time-box this so just left it here as it’s still an improvement. However, if I were to make this more generic, having the same var names for all these Observer types is useful and will make a further refactor in the future much easier.

samsymons (approved):
LGTM!

Sources/NetworkProtection/Status/ControlllerErrorMessageObserver/ControllerErrorMesssageObserver.swift
samsymons: Yeah that's a great point, I've come around on this since I posted that comment - happy to keep this as-is.

PR #462, opened 2023-08-16
SwiftLint fixes part 3

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1205234591569346/f
iOS PR: No changes needed, please drop BSK onto the iOS repo to test
macOS PR: No changes needed, please drop BSK onto the iOS repo to test
What kind of version bump will this require?: Patch, no functional changes are made

**Optional**:

Tech Design URL:
CC:

**Description**:

This PR fixes all remaining SwiftLint violations.

**Steps to test this PR**:
1. Check that CI passes
1. Check that `swiftlint .` passes when you have this branch checked out locally, as it will lint all files

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


.swiftlint.yml
```diff
@@ -1,4 +1,5 @@
 disabled_rules:
+  - large_tuple
```
samsymons, line 2: Might remove this since I don't believe our other repos have it, and just address it in areas where it comes up.

Sources/SyncDataProviders/Credentials/CredentialsProvider.swift
```diff
@@ -145,7 +168,6 @@ public final class CredentialsProvider: DataProvider {
                         }
                     default:
                         saveError = error
-                        break
```
samsymons, line 148: Interestingly, this one tripped SwiftLint's unnecessary break rule, but the other cases didn't. Is it because this is the `default` block?

Sources/BrowserServicesKit/PrivacyConfig/PrivacyConfigurationManager.swift
```diff
@@ -87,6 +87,7 @@ public class PrivacyConfigurationManager: PrivacyConfigurationManaging {
                 data = embedded
             } else {
                 let jsonData = embeddedDataProvider.embeddedData
+                // swiftlint:disable:next force_try
```
samsymons, line 90: Doing my best to make no actual logic changes where possible - my rationale here is that this has been a force unwrap for a very long time, so it's safer to just disable the lint warning rather than mess with it.

Sources/Common/TimeoutError.swift
```diff
@@ -27,8 +27,8 @@ public struct TimeoutError: Error, LocalizedError, CustomDebugStringConvertible
     public let file: StaticString
     public let line: UInt
 
-    // swiftlint:disable:next line_length
     public var errorDescription: String? {
+        // swiftlint:disable:next line_length
```
samsymons, line 31: Would rather disable this than mess with wrapping ultra long strings like this one

Sources/BrowserServicesKit/Suggestions/SuggestionProcessing.swift
```diff
@@ -80,7 +80,6 @@ final class SuggestionProcessing {
     // MARK: - DuckDuckGo Suggestions
 
     private func duckDuckGoSuggestions(from result: APIResult?) throws -> [Suggestion]? {
-        // TODO add query as duckduckgo suggestion
```
THISISDINOSAUR, line 83: Should I be concerned this never got done?

Tests/.swiftlint.yml
```diff
@@ -3,7 +3,9 @@ disabled_rules:
   - force_try
   - file_length
   - function_body_length
+  - function_parameter_count
   - identifier_name
+  - large_tuple
```
THISISDINOSAUR, line 8: Whatever you do with the other one, I guess we should do the same here

Tests/NavigationTests/NavigationAuthChallengeTests.swift
```diff
@@ -295,3 +294,7 @@ class NavigationAuthChallengeTests: DistributedNavigationDelegateTestsBase {
     }
 
 }
+
+// swiftlint:enable unused_closure_parameter
+// swiftlint:enable opening_brace
+// swiftlint:enable trailing_comma
```
THISISDINOSAUR, line 300: We're generally quite bad at reenabling these rules, especially in tests, does swift lint catch them itself?

.swiftlint.yml
samsymons: Gonna pull this one out for now :+1:

Sources/BrowserServicesKit/Suggestions/SuggestionProcessing.swift
samsymons: I had the same thought, I checked when it was added and it looked like it had existed for 2+ years. @tomasstrba do you know if this is still applicable?

Tests/NavigationTests/NavigationAuthChallengeTests.swift
samsymons: It does now tell us when you blanket disable a rule and don't re-enable it, yep - it's quite nice actually

Sources/BrowserServicesKit/LinkProtection/LinkCleaner.swift
```diff
@@ -33,10 +33,8 @@ public class LinkCleaner {
     
     public func ampFormat(matching url: URL) -> String? {
         let ampFormats = TrackingLinkSettings(fromConfig: privacyConfig).ampLinkFormats
-        for format in ampFormats {
-            if url.absoluteString.matches(pattern: format) {
-                return format
-            }
+        for format in ampFormats where url.absoluteString.matches(pattern: format) {
```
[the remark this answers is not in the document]
samsymons: You know, I actually don't know: and the uncertainty is enough that I am tempted to just leave this code entirely alone, and disable the warning around it instead.

THISISDINOSAUR (approved):

Sources/BrowserServicesKit/Suggestions/SuggestionProcessing.swift
samsymons: Created an Asana task to track this so that the PR isn't blocked

tomasstrba: This can be removed (even from Asana) because it's just a forgotten comment. The point behind it was to create a suggestion from the query like the Firefox does but we don't need that.

Screenshot 2023-08-18 at 8 57 31 AM

PR #476, opened 2023-08-24
Fix AdClickAttributionLogic leaking NavigationResponse completionBlock

mallexxx:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1199230911884351/1205322007351871/f
iOS PR: not needed
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1538

**Description**:
- Fixes async `onProvisionalNavigation` leaking completionBlocks on Navigation Task cancellation
- Added more debug info to `DistributedNavigationDelegate.checkLongDecisionMaking`

**Steps to test this PR**:
1. Set random Integration Tests execution order in Xcode scheme->Tests->Integration Tests Options
2. Run the Integration Tests, validate there‘s no `checkLongDecisionMaking` assertion


Sources/BrowserServicesKit/ContentBlocking/AdClickAttribution/AdClickAttributionLogic.swift
```diff
@@ -97,6 +97,7 @@ public class AdClickAttributionLogic {
         case .noAttribution:
             self.state = state
         case .preparingAttribution(let vendor, let info, _):
+            cancelAttributedRulesCompilationIfNeeded()
```
bwaresiak, line 100: `applyInheritedAttribution` should be called only when creating new tab from the context of the attributed tab - it should have no existing state to "cancel" at that point, so if it has, this feels like a possible problem.

Sources/BrowserServicesKit/ContentBlocking/AdClickAttribution/AdClickAttributionLogic.swift
```diff
@@ -196,12 +202,14 @@ public class AdClickAttributionLogic {
                disableAttribution()
            } else if tld.eTLDplus1(host) == vendor {
                os_log(.debug, log: log, "Refreshing navigational duration for attribution")
+               cancelAttributedRulesCompilationIfNeeded()
```
bwaresiak, line 205: Would that do anything given guard in line 179?

Sources/BrowserServicesKit/ContentBlocking/AdClickAttribution/AdClickAttributionLogic.swift
```diff
@@ -196,12 +202,14 @@ public class AdClickAttributionLogic {
...
                state = .activeAttribution(vendor: vendor,
                                           session: SessionInfo(start: session.attributionStartedAt),
                                           rules: rules)
            }
         } else if tld.eTLDplus1(host) != vendor {
             os_log(.debug, log: log, "Leaving attribution context")
+            cancelAttributedRulesCompilationIfNeeded()
```
bwaresiak, line 212: Same question as above.

Sources/BrowserServicesKit/ContentBlocking/AdClickAttribution/AdClickAttributionLogic.swift
```diff
@@ -216,8 +224,9 @@ public class AdClickAttributionLogic {
         eventReporting?.fire(.adAttributionActive)
         registerFirstActivity = false
     }
-    
+
     private func disableAttribution() {
+        cancelAttributedRulesCompilationIfNeeded()
```
bwaresiak, line 229: This is called only if state is active, so I don't think it does anything either.

Sources/Navigation/DistributedNavigationDelegate.swift
```diff
@@ -111,7 +111,8 @@ private extension DistributedNavigationDelegate {
 #endif
 
     /// continues until first non-nil Navigation Responder decision and returned to the `completion` callback
-    func makeAsyncDecision<T>(boundToLifetimeOf webView: WKWebView,
+    func makeAsyncDecision<T>(for actionDebugInfo: some CustomDebugStringConvertible,
+                              boundToLifetimeOf webView: WKWebView,
```
bwaresiak, line 115: Didn't we agree to disable this kind of warning?

mallexxx: As I remember we agreed to make it less intrusive (to stop debugger and not cause assertion) and provide more info. This won't affect CI runners as there's no debugger but may help detecting some issues

PR #480, opened 2023-08-28
Use unencrypted password for deduplicated Credentials objects

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/1199230911884351/1205352938100234/f
iOS PR: https://github.com/duckduckgo/iOS/pull/1959
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1554

**Description**:
Credentials deduplication code looks for an existing credentials entity in the Secure Vault,
then returns it to Sync code where its title is updated, and then the credentials entity is passed
back to Secure Vault to be stored. The storing function expects unencrypted password
(such as when passed from the UI) and does encryption before storing.

The issue here was that we were passing an already encrypted password (just retrieved
from Secure Vault) in which case the encryption function failed to encrypt the value, returning nil.

This patch fixes the problem on the Sync Data Provider end (to feed an unencrypted password
to Secure Vault for saving) and also adds an assertion failure to AutofillSecureVault when an encrypted
password is passed to encryptPassword method.

**Steps to test this PR**:
Refer to macOS PR for testing steps.

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/AutofillSecureVault.swift
+++ b/Sources/BrowserServicesKit/SecureVault/AutofillSecureVault.swift
@@ -336,6 +336,9 @@ public class DefaultAutofillSecureVault<T: AutofillDatabaseProvider>: AutofillSe
                                  key l2Key: Data? = nil,
                                  salt: Data? = nil) throws -> SecureVaultModels.WebsiteCredentials {
         do {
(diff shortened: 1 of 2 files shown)


graeme (approved):
Tested and code looks good 👍

PR #505, opened 2023-09-18
Improves the tunnel logic to better support on-demand

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/0/1205518995339759/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2016
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1631

## Description

Improves the tunnel logic to better support on-demand

## Steps to test this PR

See the integration PRs for testing instructions.

---
###### Internal references:


Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -208,7 +227,7 @@ final class NetworkProtectionConnectionTester {
 
     // MARK: - Testing the connection
 
-    func testConnection() {
+    func testConnection(isStartupTest: Bool) async throws {
```
diegoreymendez, line 230: Turning this async so we can do immediate calls to.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -426,26 +440,42 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     open override func startTunnel(options: [String: NSObject]?, completionHandler: @escaping (Error?) -> Void) {
         connectionStatus = .connecting
-        tunnelHealth.isHavingConnectivityIssues = false
-        controllerErrorStore.lastErrorMessage = nil
```
diegoreymendez, line 430: When starting NetP on-demand we no longer clear these, since we consider there to be an ongoing issue.

The new logic is implemented in `resetIssueStateOnTunnelStart(_:)` which is called a few lines below.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -426,26 +440,42 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+
+        os_log("Will load options\n%{public}@", log: .networkProtection, String(describing: options))
+        let startupOptions = StartupOptions(options: options ?? [:], log: .networkProtection)
+
+        resetIssueStateOnTunnelStart(startupOptions)
+
+        let startTime = DispatchTime.now()
```
diegoreymendez, line 449: Used to allow some time before failing, if we're starting with on-demand.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -426,26 +440,42 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
 
-            completionHandler(error)
+                completionHandler(error)
+            }
+
+            if startupOptions.startupMethod == .automaticOnDemand {
+                DispatchQueue.main.asyncAfter(deadline: startTime + DispatchTimeInterval.seconds(10), execute: handler)
+            } else {
+                handler()
+            }
```
diegoreymendez, line 476: This is the gist of the changes above... we're executing a completion handler immediately or with a certain delay depending on the circumstances.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -359,9 +362,9 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
     }
 
     private func load(options: StartupOptions) throws {
-        isConnectionTesterEnabled = options.enableTester
```
diegoreymendez, line 362: This is no longer good since we now cancel the tunnel, and on-demand restarts it but the option won't be present in the options dictionary.

I made some change so that the option is persisted across launches, unless explicitly overridden.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -37,6 +37,7 @@ final class NetworkProtectionConnectionTester {
 
     enum TesterError: Error {
         case couldNotFindInterface(named: String)
+        case connectionTestFailed
```
diegoreymendez, line 40: When starting the VPN on demand, we now do an immediate test on startup.  This is the error if the test fails.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -93,11 +94,11 @@ final class NetworkProtectionConnectionTester {
     // MARK: - Test result handling
 
     private var failureCount = 0
-    private let resultHandler: @MainActor (Result) -> Void
+    private let resultHandler: @MainActor (Result, Bool) -> Void
```
diegoreymendez, line 97: The bool is for letting the callback now if this is the startup test done for an on-demand startup.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -113,40 +114,37 @@ final class NetworkProtectionConnectionTester {
...
         await stopScheduledTimer()
     }
 
-    /// Run the test right now and schedule the next one regularly.
-    /// 
-    func testImmediately() async {
-        await stopScheduledTimer()
-        testConnection()
-        await scheduleTimer()
-    }
```
diegoreymendez, line 140: Removed since it's not used anywhere.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -169,23 +167,44 @@ final class NetworkProtectionConnectionTester {
...
+        await timerRunCoordinator.start()
+
+        if testImmediately {
+            do {
+                try await testConnection(isStartupTest: true)
+            } catch {
+                os_log("Rethrowing exception", log: log)
+                throw error
+            }
+        }
```
diegoreymendez, line 182: We test the connection before setting the `DispatchTimer` up, as the timer doesn't like being created but not run, and crashes.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -222,29 +241,30 @@ final class NetworkProtectionConnectionTester {
         let localParameters = NWParameters.tcp
         localParameters.prohibitedInterfaces = [tunnelInterface]
 
-        Task {
```
diegoreymendez, line 225: The diff below looks a bit ugly, but for the most part is just a tab-level change.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
samsymons: That's probably okay, I was thinking we could perhaps make a typealias out of it or evenprovide a struct to encapsulate everything cleanly (since looking at this now, it's not clear what the Bool is for), but I'm not going to block on it.

Perhaps even just a doc comment is the way to go.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -113,40 +114,37 @@ final class NetworkProtectionConnectionTester {
...
-        os_log("🟢 Starting connection tester", log: log, type: .debug)
+        os_log("🟢 Starting connection tester (testImmediately: %{public}@)", log: log, String(reflecting: testImmediately))
         let tunnelInterface = try await networkInterface(forInterfaceNamed: tunnelIfName)
         self.tunnelInterface = tunnelInterface
 
-        await scheduleTimer()
+        do {
+            try await scheduleTimer(testImmediately: testImmediately)
+        } catch {
+            os_log("🔴 Stopping connection tester early", log: log)
```
samsymons, line 130: This could use `type: .error`

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -169,23 +167,44 @@ final class NetworkProtectionConnectionTester {
...
+    private func scheduleTimer(testImmediately: Bool) async throws {
         await stopScheduledTimer()
 
+        await timerRunCoordinator.start()
+
+        if testImmediately {
+            do {
+                try await testConnection(isStartupTest: true)
+            } catch {
+                os_log("Rethrowing exception", log: log)
```
samsymons, line 179: I'd be inclined to remove this log, and let the caller who has to catch this error do the logging. Or we can just make it more clear like "Caught exception when testing connection", since this would be confusing to see on its own in the logs.

Sources/NetworkProtection/WireGuardKit/WireGuardAdapter.swift
```diff
@@ -312,12 +313,12 @@ public class WireGuardAdapter {
                 // Tell the system that the tunnel is going to reconnect using new WireGuard
                 // configuration.
                 // This will broadcast the `NEVPNStatusDidChange` notification to the GUI process.
-                self.packetTunnelProvider?.reasserting = true
+                //self.packetTunnelProvider?.reasserting = true
```
samsymons, line 316: Any reason to keep these commented out rather than removing?

samsymons (approved):
This is working well for me, left a few comments but feel free to merge once those are resolved.

PR #546, opened 2023-10-31
Fix syncing empty favorites folders

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/414235014887631/1205843304285892/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2121
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1791

**Description**:
Update bookmarks sync response handler to actually process favorites
also when an empty folder is received from the server.

**Steps to test this PR**:
1. Connect 2 devices in Sync. Populate favorites.
2. Remove all favorites on one device (it's easiest on macOS).
3. Verify that all favorites are removed from another device after sync completes.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/SyncDataProviders/Bookmarks/internal/BookmarksResponseHandler.swift
```diff
@@ -113,33 +113,34 @@ final class BookmarksResponseHandler {
         }
         try processOrphanedBookmarks()
 
-        // populate favorites
-        if !favoritesUUIDs.isEmpty {
-            guard let favoritesFolder = BookmarkUtils.fetchFavoritesFolder(context) else {
-                // Error - unable to process favorites
-                return
-            }
+        processReceivedFavorites()
```
ayoy, line 116: Just moving favorites handling to a separate function.

Sources/SyncDataProviders/Bookmarks/internal/BookmarksResponseHandler.swift
```diff
@@ -113,33 +113,38 @@ final class BookmarksResponseHandler {
...
-            }
+    // MARK: - Private
 
-            favoritesUUIDs.forEach { uuid in
-                if let bookmark = entitiesByUUID[uuid] {
-                    bookmark.removeFromFavorites()
-                    bookmark.addToFavorites(favoritesRoot: favoritesFolder)
-                }
+    private func processReceivedFavorites() {
+        guard let favoritesUUIDs else {
```
bwaresiak, line 122: I wonder - is it possible if we don't receive fav root folder in the initial sync? Shall we fire a pixel for `favoritesUUIDs == nil && shouldDeduplicateEntities` ? Also in such case do we retain modified at for favoritesFolderID?

Sources/SyncDataProviders/Bookmarks/internal/BookmarksResponseHandler.swift
```diff
@@ -57,7 +57,7 @@ final class BookmarksResponseHandler {
         var allUUIDs: Set<String> = []
         var childrenToParents: [String: String] = [:]
         var parentFoldersToChildren: [String: [String]] = [:]
-        var favoritesUUIDs: [String] = []
+        var favoritesUUIDs: [String]? = nil
```
bwaresiak, line 61: There seem to be some linter warnings.

Sources/SyncDataProviders/Bookmarks/internal/BookmarksResponseHandler.swift
ayoy: * It's unlikely that we don't receive favorites folder(s) in the initial sync. I think it can happen when the first device  in a Sync account fails to do the first sync and then a new device joins the account. Otherwise, the first device is supposed to send all its data, including favorites folders, and backend is supposed to send all account data in the initial sync to new devices joining the account.
  * Can you think of any other scenario (besides backend going rogue)?
* Tracking this with pixel is to me equivalent to tracking e.g. root folder missing in the initial sync payload. We're not doing it at the moment and if it happens, we're screwed (the devices will overwrite their data eventually).
  * Perhaps we should also track failed first sync on the account?
* We don't retain modified timestamps here (there's no custom logic for it). Favorites folder missing from the server response doesn't indicate that the server didn't process the sent object. But if it's initial sync, then nothing was sent so nothing would have their modified timestamp cleared, right?
* In any case, I don't think we should update this very PR with a different handling of data missing from Sync response, but rather tackle this as a separate Asana task, wdyt?

Sources/SyncDataProviders/Bookmarks/internal/BookmarksResponseHandler.swift
```diff
@@ -113,33 +113,38 @@ final class BookmarksResponseHandler {
...
+        }
+
+        guard let favoritesFolder = BookmarkUtils.fetchFavoritesFolder(context) else {
+            // Error - unable to process favorites
+            return
+        }
+
+        // For non-first sync we rely fully on the server response
+        if !shouldDeduplicateEntities {
+            favoritesFolder.favoritesArray.forEach { $0.removeFromFavorites() }
```
bwaresiak, line 133: As per call - we will need to validate local timestamp here.

bwaresiak (approved):
LGTM!

bwaresiak (approved):
LGTM

Sources/SyncDataProviders/Bookmarks/internal/BookmarksResponseHandler.swift
ayoy: Added a task [here](https://app.asana.com/0/1201493110486074/1205896860658963/f).

PR #560, opened 2023-11-11
Attempts to fix a crasher with BloomFilterWrapper in macOS 11.x

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/0/1205931560838491/f
iOS PR: 
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1839

## Description

Attempts to provide a fix to a crash that's happening in macOS 11.x systems due to BloomFilterWrapper.

## Testing

Just make sure everything builds fine.  There are no logical changes.

---
###### Internal references:

### DIFF
--- a/Sources/BloomFilterObjC/include/module.private.modulemap
+++ b/Sources/BloomFilterObjC/include/module.private.modulemap
@@ -1,4 +1,4 @@
-module BloomFilterWrapper {
+module BloomFilterObjC {
     header "BloomFilterObjC.h"
     export *
 }
--- a/Sources/BloomFilterWrapper/BloomFilterWrapper.swift
+++ b/Sources/BloomFilterWrapper/BloomFilterWrapper.swift
@@ -22,18 +22,22 @@ import Foundation
 public final class BloomFilterWrapper {
     private let bloomFilter: BloomFilterObjC
 
+    @inline(never)
     public init(fromPath path: String, withBitCount bitCount: Int32, andTotalItems totalItems: Int32) {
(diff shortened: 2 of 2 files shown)


Sources/BloomFilterWrapper/BloomFilterWrapper.swift
```diff
@@ -22,18 +22,22 @@ import Foundation
 public final class BloomFilterWrapper {
     private let bloomFilter: BloomFilterObjC
 
+    @inline(never)
```
diegoreymendez, line 25: One thing I noticed in the crash logs is that these methods seem to be optimized marked as `compiler-generated`.  This normally means those methods have been optimized... possibly inlined.

Inlined methods are basically compiled in the locations where they're called, instead of kept as stand-alone methods.  I believe this may clash with `@_implementationOnly` since only this module has access to `BloomFilterObjC`, so an inlined method may fail to reference it.

By adding `@inline(never)` here I'm hoping to force these methods to be compiled as regular methods where they are.

Ref: https://embrace.io/blog/compiler-generated-ios-crashes/

Sources/BloomFilterObjC/include/module.private.modulemap
```diff
@@ -1,4 +1,4 @@
-module BloomFilterWrapper {
+module BloomFilterObjC {
```
diegoreymendez, line 1: This was clearly not right, although I'm not sure it could've caused trouble.

diegoreymendez:
Closing this as it didn't resolve the issue.

PR #565, opened 2023-11-15
Adds support for managing VPN settings.

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/0/1205958731729757/f

iOS PR: https://github.com/duckduckgo/iOS/pull/2165
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1858

## Description

Adds support for the settings we'll be showing in our macOS Settings pane.

## Testing

Look at the platform specific PRs for testing instructions.

---
###### Internal references:


Sources/NetworkProtection/AppLaunching.swift
```diff
@@ -25,6 +25,7 @@ public enum AppLaunchCommand: Codable {
     case justOpen
     case shareFeedback
     case showStatus
+    case showSettings
```
diegoreymendez, line 28: New launch option to direct the user to VPN Settings.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -97,7 +97,8 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             guard connectionStatus != oldValue else {
                 return
             }
-            if case .connected = connectionStatus {
+            if case .connected = connectionStatus,
+               self.settings.notifyStatusChanges {
```
diegoreymendez, line 101: We're now checking if notifications are enabled before presenting them.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -583,9 +586,13 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             }
 
             Task { [weak self] in
-                await self?.handleAdapterStopped()
-                if case .superceded = reason {
-                    self?.notificationsPresenter.showSupersededNotification()
+                if let self {
```
diegoreymendez, line 589: Just unwrapping self.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -823,11 +830,15 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+        case .setConnectOnLogin,
+                .setIncludeAllNetworks,
                 .setEnforceRoutes,
                 .setExcludeLocalNetworks,
+                .setNotifyStatusChanges,
                 .setRegistrationKeyValidity,
-                .setSelectedEnvironment:
+                .setSelectedEnvironment,
+                .setShowInMenuBar,
+                .setShowVPNSettings:
```
diegoreymendez, line 841: Added several new options that don't need handling in the extension.

Still it's cool that the settings mechanism allows us to handle these if anything changed.  Like for example if eventually the VPN settings could be directly changed from the extension.  Or if we wanted to fire a pixel.

Sources/NetworkProtection/Settings/TunnelSettings.swift
```diff
@@ -275,6 +330,54 @@ public final class TunnelSettings {
...
+
+    public var notifyStatusChanges: Bool {
+        get {
+            defaults.networkProtectionNotifyStatusChanges
+        }
+
+        set {
+            defaults.networkProtectionNotifyStatusChanges = newValue
+        }
+    }
```
samsymons, line 379: I wonder if a Macro could be used to save space with this boilerplate code. (I wouldn't actually investigate that, since we can't use Xcode 15 everywhere yet - just thinking out loud.)

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: Does this make `NetworkProtectionNotificationsPresenterTogglableDecorator` within the iOS implementation redundant?

diegoreymendez: I was surprised to see iOS not using `TunnelSettings` for a newly introduced setting to be honest, as I had shared this class a while back as a mechanism to add settings that can be supported across platforms.

This is in part why I would like some of these projects to be run across both platforms simultaneously.

Sources/NetworkProtection/Settings/TunnelSettings.swift
diegoreymendez: Yes, I have considered that.  I also considered using a protocol, but there are going to be some unique challenges.

The way I currently see `TunnelSettings` is that it's showing us through it's organic growth that it may need a refactoring to remove some of the code duplication.

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: It doesn't make the decorator unnecessary, though.  I can switch to using the decorator for macOS too.

I'll push that.

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: Nice :+1:

samsymons (approved):

PR #575, opened 2023-11-24
Improve handling lists in Sync

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1205884549503153/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2192
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1886

**Optional**:

Tech Design URL:
CC:

**Description**:
Improve handling of lists in syncable models' payloads (which currently is limited to bookmarks
folders and favorites folders children). This change updates how lists are sent to Sync in PATCH
payloads by including current state as well as the set of removed and inserted items.
To allow for this, last sync payload is kept for all bookmarks folders in
`lastChildrenPayloadReceivedFromSync` property added on BookmarkEntity entity.
If the payload is not available (e.g. first sync), the legacy mechanism is used (i.e. sending only
the current contents of the list). The schema of a Sync response payload does not change so
the implementation handles it too.

**Steps to test this PR**:
1. Run the client app (iOS and/or macOS)
2. Import/populate bookmarks and enable Sync.
3. Move bookmarks across folders and verify that changes are properly synced between devices.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/BookmarksTestsUtils/BookmarkTree.swift
ayoy, line 1: I updated `BookmarkTree` to support setting and verifying `lastChildrenArrayReceivedFromSync`. It's verified by default (and I updated a bunch of test to verify it), but can be disabled if it doesn't make sense to verify it.

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
```diff
@@ -58,6 +58,7 @@ public final class BookmarksProvider: DataProvider {
             let bookmarks = (try? context.fetch(fetchRequest)) ?? []
             for bookmark in bookmarks {
                 bookmark.modifiedAt = Date()
+                bookmark.lastChildrenArrayReceivedFromSync = nil
```
ayoy, line 61: Nullify last payload for all bookmarks before first sync.

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
```diff
@@ -154,7 +155,7 @@ public final class BookmarksProvider: DataProvider {
         if sent.isEmpty {
             return []
         }
-        let identifiers = sent.compactMap { SyncableBookmarkAdapter(syncable: $0).uuid }
+        let identifiers = sent.compactMap { SyncableBookmarkAdapter(syncable: $0, isPatchRequestPayload: true).uuid }
```
ayoy, line 158: SyncableBookmarkAdapter gets a boolean parameter to decide on how the `children` array should be read (because Sync response uses the old format).

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
```diff
@@ -167,6 +168,13 @@ public final class BookmarksProvider: DataProvider {
             if bookmark.isPendingDeletion, !isLocalChangeRejectedBySync {
                 context.delete(bookmark)
             } else {
+                if !isLocalChangeRejectedBySync, bookmark.isFolder {
```
ayoy, line 171: This is inside `cleanUpSentItems` which is called before processing the response. If a bookmark was rejected by sync, it comes with the response, will be processed there and its sync payload will be updated then.

Sources/Bookmarks/BookmarksModel.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>BookmarksModel 4.xcdatamodel</string>
+	<string>BookmarksModel 5.xcdatamodel</string>
```
bwaresiak, line 6: Note that I'd expect this to trigger (again) the fix for migrating the favorites on the platforms - I don't think it's harmful, just something to remember.

Sources/SyncDataProviders/Bookmarks/internal/SyncableBookmarkAdapter.swift
```diff
@@ -51,10 +53,23 @@ struct SyncableBookmarkAdapter {
     }
 
     var children: [String] {
-        guard let folder = syncable.payload["folder"] as? [String: Any], let folderChildren = folder["children"] as? [String] else {
+        guard let folder = syncable.payload["folder"] as? [String: Any] else {
             return []
         }
-        return folderChildren
+
+        if isPatchRequestPayload,
```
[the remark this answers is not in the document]
ayoy: I removed it 👍

SabrinaTardio (commented):
I tested a few cases 
for example adding something to the same folder to 2 connected device closely together.
adding and removing form the 2 devices
changing folders etc.
Seems to work well, everything is synced correctly.

I have a couple of questions (for my understanding)

I don’t have a lot of context on this codebase but changes seem to make sense. If Bartek had a look as well (which seems like he did) I am happy to approve.

Sources/SyncDataProviders/Bookmarks/BookmarksProvider.swift
ayoy: `isPendingDeletion` is required to handle "soft deletion". Objects that are deleted from the UI aren't removed from the database, but instead are marked as pending deletion and filtered out in the UI. They are only deleted after the deletion is synced to the server.

SabrinaTardio (approved):
LGTM

PR #613, opened 2023-12-22
Update latency & tunnel failure monitors implementation

quanganhdo:
**Required**:

Task/Issue URL: https://app.asana.com/0/0/1206209663744590/f and https://app.asana.com/0/0/1206212872939262/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2290
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2005

**Description**:

Updates the monitor implementation to use Task to address situations when timer is launched multiple times. This affects macOS where it may have caused pixel anomalies due.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
@@ -83,28 +83,28 @@ public actor NetworkProtectionLatencyMonitor {
     @MainActor
     private var ignoreThreshold = false
 
-    private let serverIP: () -> IPv4Address?
-
-    private let log: OSLog
+    @MainActor
+    private(set) var serverIP: IPv4Address?
 
     // MARK: - Init & deinit
 
-    init(serverIP: @escaping () -> IPv4Address?, log: OSLog) {
-        self.serverIP = serverIP
-        self.log = log
-
+    init() {
         os_log("[+] %{public}@", log: .networkProtectionMemoryLog, type: .debug, String(describing: self))
     }
 
     deinit {
+        task?.cancel()
+
         os_log("[-] %{public}@", log: .networkProtectionMemoryLog, type: .debug, String(describing: self))
     }
 
     // MARK: - Start/Stop monitoring
 
     @MainActor
-    public func start(callback: @escaping (Result) -> Void) {
-        os_log("⚫️ Starting latency monitor", log: log)
+    public func start(serverIP: IPv4Address, callback: @escaping (Result) -> Void) {
+        os_log("⚫️ Starting latency monitor", log: .networkProtectionPixel)
+
+        self.serverIP = serverIP
 
         latencyCancellable = latencySubject.eraseToAnyPublisher()
             .scan(ExponentialGeometricAverage()) { measurements, latency in
@@ -136,7 +136,7 @@ public actor NetworkProtectionLatencyMonitor {
 
     @MainActor
     public func stop() {
-        os_log("⚫️ Stopping latency monitor", log: log)
+        os_log("⚫️ Stopping latency monitor", log: .networkProtectionPixel)
 
         latencyCancellable = nil
         task = nil
@@ -146,7 +146,7 @@ public actor NetworkProtectionLatencyMonitor {
 
     @MainActor
     public func measureLatency() async {
-        guard let serverIP = serverIP() else {
+        guard let serverIP else {
             latencySubject.send(Self.unknownLatency)
             return
         }
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
@@ -59,11 +59,7 @@ public actor NetworkProtectionLatencyMonitor {
 
     private static let unknownLatency: TimeInterval = -1
 
-    public nonisolated var publisher: AnyPublisher<Result, Never> {
-        subject.eraseToAnyPublisher()
-    }
-    private let subject = PassthroughSubject<Result, Never>()
-
+    @MainActor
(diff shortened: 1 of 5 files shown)


diegoreymendez:
@quanganhdo - I'm starting the review for this PR.  Could I kindly ask to always include a description of the changes and a list of manual tests, ideally with 100% coverage of the changes?  Thanks!

Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
```diff
@@ -21,7 +21,7 @@ import Network
 import Common
 import Combine
 
-final public class NetworkProtectionLatencyMonitor {
+public actor NetworkProtectionLatencyMonitor {
```
diegoreymendez, line 24: Making this an actor already ensures that its methods are executed sequentially, so it doesn't make sense for this to be an actor and its methods and properties to be declared as `@MainActor`.

Why not do without `@MainActor` entirely?

Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
```diff
@@ -59,95 +59,60 @@ final public class NetworkProtectionLatencyMonitor {
...
     // MARK: - Start/Stop monitoring
 
-    public func start() async throws {
-        guard await !timerRunCoordinator.isRunning else {
-            os_log("Will not start the latency monitor as it's already running", log: log)
-            return
-        }
+    @MainActor
+    public func start(serverIP: IPv4Address, callback: @escaping (Result) -> Void) {
+        os_log("⚫️ Starting latency monitor", log: .networkProtectionPixel)
```
diegoreymendez, line 105: Why are we logging to the pixel log?

Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
```diff
@@ -39,137 +39,64 @@ final public class NetworkProtectionTunnelFailureMonitor {
...
 
-    private var timer: DispatchSourceTimer?
-    private let timerRunCoordinator = TimerRunCoordinator()
-    private let timerQueue: DispatchQueue
+    @MainActor
+    var isStarted: Bool {
+        task?.isCancelled == false
+    }
 
     private let tunnelProvider: PacketTunnelProvider
```
diegoreymendez, line 54: Should this be `weak` to prevent a retain cycle?

diegoreymendez (changes requested):
I currently can test the macOS build as it's failing.  Added some feedback.

Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
```diff
@@ -159,71 +124,29 @@ final public class NetworkProtectionLatencyMonitor {
...
+        task = nil
     }
 
     // MARK: - Latency monitor
 
     @MainActor
     public func measureLatency() async {
-        guard let serverIP = serverIP() else {
+        guard let serverIP else {
             latencySubject.send(Self.unknownLatency)
```
diegoreymendez, line 150: If this happens the state would be badly broken, and reporting unkown latency would silently hide the issue.

Can we instead set `serverIP` in the `init` method so it's never optional?  It would clean up the code a bit and it would help us remove these lines which would make things clearer.

quanganhdo: `serverIP` is taken from `lastSelectedServerInfo?.ipv4` so it might not be known at initialization time. It also needs updating when server location changes. At call site, there’s a check for the ip to be present:

```
guard let ip = lastSelectedServerInfo?.ipv4 else {
     latencyMonitor.stop()
     return
}
```

Should I use an `assertionFailure` instead inside the `measureLatency()` method above to indicate a programmer error? `start(serverIP:callback:)` requires a non-nil ip anyway.

Lmk what you think.

diegoreymendez: I'd consider adding the server IP as a non optional param to `measureLatency(to address: IPv4Address)`.  That way we don't need an optional property.

Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
diegoreymendez: This one's important as it may also explain why the timer wasn't being released before.

diegoreymendez (changes requested):

diegoreymendez:
This seems to work for me.  That said I'm getting a lot of these:

```
⚫️ Ping error: The operation couldn’t be completed. (NetworkProtection.Pinger.PingError error 2.)
```

Any idea why that'd be the case?

Sources/NetworkProtection/Diagnostics/NetworkProtectionLatencyMonitor.swift
```diff
@@ -55,196 +55,107 @@ final public class NetworkProtectionLatencyMonitor {
...
-            lock.lock(); defer { lock.unlock() }
-            return _ignoreThreshold
-        }
-        set {
-            lock.lock()
-            self._ignoreThreshold = newValue
-            lock.unlock()
-        }
-    }
+    private(set) var serverIP: IPv4Address?
```
diegoreymendez, line 78: This should be removed.

diegoreymendez (approved):
I'm approving, but please take a moment to look at my last comment since there's some cleanup to do before merging.

PR #628, opened 2024-01-20
Fix tunnel monitor check

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/414235014887631/1206397132422633/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2361
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2096

**Description**:

This PR improves the tunnel monitor check by ignoring handshake values that are equal to zero, avoiding false failure pings.

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -994,9 +994,11 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             connectionStatus = .connected(connectedDate: Date())
         }
 
-        guard !isKeyExpired else {
-            await rekey()
-            return
+        if !settings.disableRekeying {
+            guard !isKeyExpired else {
+                await rekey()
+                return
+            }
         }
 
         os_log("🔵 Tunnel interface is %{public}@", log: .networkProtection, type: .info, adapter.interfaceName ?? "unknown")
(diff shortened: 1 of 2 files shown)


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -994,9 +994,11 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             connectionStatus = .connected(connectedDate: Date())
         }
 
-        guard !isKeyExpired else {
-            await rekey()
-            return
+        if !settings.disableRekeying {
```
diegoreymendez, line 997: Please bear with me for a sec as I'm confused by this.  It certainly looks right to do this if there's a setting, but I'm lacking context about why this was added.

My question is two-fold and it's mostly to catch up with our plans:
1. Why was this added to begin with?
2. What's the plan for rekeying that makes us want to disable it optionally?

diegoreymendez (approved):
Works well for me.

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: This was a debug option setting that I added last year when I was talking with Aitor about the effects of removing rekeying - it's not used for prod users in any way. There's a discussion task in Asana hashing out the merits of removing rekeying, but I forget where it got to.

This change here is just a bugfix that tripped me up while testing this feature - I had the setting to disable rekeying turned on, which was causing the app to always get to this `isKeyExpired` check and never start the tunnel monitor.

PR #633, opened 2024-01-24
Display different copy in alerts when a flag is enabled in a newer version

SabrinaTardio:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1206310394453442/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2374
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2114

**Description**: Shows a different copy in alerts when a sync flag is enabled in a newer version and the user needs to update the app

**Steps to test this PR**:
1. in the PrivacyConfig file increase the min version for the sync feature and each subfeature one at a time.
2. In both sync the enabled and disabled sync setting version (if visible) check that the right alert is shown (asking user to upgrade)
3. Disable completely the flags one at a time,  and check and go to the sync settings and check the alert is the generic sync is unavailable one

<!—
Before submitting a PR, please ensure you have tested the combinations you expect the reviewer to test, then delete configurations you *know* do not need explicit testing.

Using a simulator where a physical device is unavailable is acceptable.
—>

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

—
###### Internal references:

### DIFF
--- a/Sources/DDGSync/SyncFeatureFlags.swift
+++ b/Sources/DDGSync/SyncFeatureFlags.swift
@@ -24,6 +24,7 @@ import Foundation
  */
 public struct SyncFeatureFlags: OptionSet {
     public let rawValue: Int
+    public var unavailableReason: PrivacyConfigurationFeatureDisabledReason?
 
     public init(rawValue: Int) {
         self.rawValue = rawValue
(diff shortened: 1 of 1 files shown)


Sources/DDGSync/SyncFeatureFlags.swift
```diff
@@ -24,6 +24,7 @@ import Foundation
  */
 public struct SyncFeatureFlags: OptionSet {
     public let rawValue: Int
+    public var unavailableReason: PrivacyConfigurationFeatureDisabledReason?
```
ayoy, line 27: This should be a `public private(set) var`, so that it can't be modified from outside.

Sources/DDGSync/SyncFeatureFlags.swift
```diff
@@ -68,20 +69,39 @@ public struct SyncFeatureFlags: OptionSet {
     // MARK: -
 
     init(privacyConfig: PrivacyConfiguration) {
-        guard privacyConfig.isEnabled(featureKey: .sync) else {
+        var offFlag: SyncSubfeature?
```
ayoy, line 72: Wouldn't calling it `disabledSubfeature` make it more clear?

Sources/DDGSync/SyncFeatureFlags.swift
ayoy, line 27: I like this approach a lot. I didn't think about it initially because I'm having this mindset that OptionSet is basically an enum so we shouldn't be adding other fields to it, besides the raw value. But since enums have associated values, why not add an extra field to an OptionSet struct if it's needed. It looks like an equivalent of an enum's associated value. 

Good job! 👏

Sources/DDGSync/SyncFeatureFlags.swift
```diff
@@ -68,20 +69,39 @@ public struct SyncFeatureFlags: OptionSet {
...
-            self = .level3AllowCreateAccount
+
+        if let offFlag {
+            let state = privacyConfig.stateFor(offFlag)
+            switch state {
+            case .enabled:
+                break
+            case .disabled(let reason):
+                unavailableReason = reason
+            }
```
ayoy, line 104: This was a bit confusing to me at the beginning and I had to follow the code closely to understand it better. Could we rewrite it as follows?

```suggestion
        if let offFlag, case .disabled(let reason) = privacyConfig.stateFor(offFlag) {
            unavailableReason = reason
        }
```

ayoy (approved):
LGTM! I left a few comments that could improve readability, but I'm only strong on the `public private(set)` thingy, otherwise it's solid and you can decide to address remaining comments or not. Will have a look at client PRs now. Thanks @SabrinaTardio!

Sources/DDGSync/SyncFeatureFlags.swift
SabrinaTardio: Yes! def should have done that

SabrinaTardio: I know what you mean lol… initially it didn’t come to me intuitively to do that… and then I thought… why not! we do it for enum too ahahah

Sources/DDGSync/SyncFeatureFlags.swift
SabrinaTardio: I personally find this less readable because is condensed but I’ll change it… cause it's probably just me 😄

PR #639, opened 2024-01-29
Improve VPN auth token storage

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/1199333091098016/1206460327917918/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2398
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2129

**Description**:

This PR improves the way the auth token is handled for the VPN:

* The token is no longer deleted and then re-added so frequently; instead, it gets updated only when required
* We no longer emit error events when an invalid invite code is entered manually as this case is expected

**Steps to test this PR**:
1. See client PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -408,8 +408,16 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-        case .set(let authToken):
-            try tokenStore.store(authToken)
+        case .set(let newAuthToken):
+            do {
+                if let currentAuthToken = try tokenStore.fetchToken(), currentAuthToken == newAuthToken {
+                    return
+                }
+            } catch {
+                // Do nothing; in this case the token store doesn't have its token yet.
+            }
```
samsymons, line 418: This check simply avoids trying to store an auth token that already exists - it should likely use `try?` instead though, since we still want to try to save the code if something goes wrong when reading it.

Sources/NetworkProtection/KeyManagement/NetworkProtectionTokenStore.swift
```diff
@@ -58,10 +58,6 @@ public final class NetworkProtectionKeychainTokenStore: NetworkProtectionTokenSt
     public func store(_ token: String) throws {
         let data = token.data(using: .utf8)!
         do {
-            try (try? keychainStore.deleteAll()) ?? {
-                // sometimes it fails from the first try: retry once
-                try keychainStore.deleteAll()
-            }()
```
samsymons, line 64: We no longer need to delete values before writing them, as the `writeData` function now updates keychain entries instead of failing to save them due to the duplicate item error.

graeme (approved):

PR #642, opened 2024-01-30
Add error codes to site breakage reports

jaceklyp:
**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1205982573174191/f
iOS PR: 
macOS PR: 

**Optional**:

Tech Design URL: https://app.asana.com/0/0/1206225370438701/f
CC: @jonathanKingston 

**Description**:
Send error and http status codes within reports

**Steps to test this PR**:
1. Load https://expired.badssl.com/
1. Send report
1. Put a breakpoint and validate that the report contains error

1. Load any site
1. Send report
1. Put a breakpoint and validate that the report contains httpErrorCodes


Sources/PrivacyDashboard/WebsiteBreakage/WebsiteBreakage.swift
```diff
@@ -155,6 +165,15 @@ public struct WebsiteBreakage {
             result["lastSentDay"] = lastSentDay
         }
 
+        if let httpStatusCode {
+            result["httpErrorCode"] = String(httpStatusCode)
+        }
+
+        if let error = error as NSError? {
+            let errorDescription = "\(error.code) - \(error.localizedDescription)"
```
federicocappelli, line 173: 1 - I assume that the format "666 - This is an error" will be used in grafana/kibana for reports, should we create a simple unit test for fixing it? we wouldn't want this to change in the future

2 - What if the error is a swift Error and not an NSError? is this if == false?

federicocappelli (changes requested):
Only one question and one suggestion

Sources/PrivacyDashboard/WebsiteBreakage/WebsiteBreakage.swift
jaceklyp: 1. I think this can be subject of change - we look at reports one by one, and do not necessarily want to sort/group by these errors. But I'm willing to add one as a follow up :)
2. From what I understand this comes from the WebKit which is written in objC and these happen to be NSErrors, it's hard to tell, because documentation really sucks :)

federicocappelli: 1. fair enough if it's only for human consumption we don't need a test
2. My only problem here is that the interface is ambiguous then, the var is ` error: Error?` but then only NSError is a valid content, what happens if in future somebody else uses that field for a normal swift error and the value simply disappear, without an error thrown? can we add an else for normal errors or change the var type to NSError?

federicocappelli: in iOS in fact that error is populated with the content `private var lastError: Error?`, nobody guarantees that error to be a NSError

federicocappelli: My only concern is that this piece of information will be lost immediately. The future developer using this code will see that accepts Error but in fact, only NSError is valid/used and a swift Error is discarded without notice.
I would just add a `if NSError else if Error`

federicocappelli (approved):
LGTM, tested, all good

PR #649, opened 2024-02-01
Check subfeature state before reading rollout data

dus7:
Asana task(s): https://app.asana.com/0/414235014887631/1206494747562366/f

iOS PR: https://github.com/duckduckgo/iOS/pull/2426
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2152

**Description:**
This fix comprises of two parts:

##### Fix subfeature state resolution logic
Rollout state is checked before the global subfeature state, resulting in reporting invalid reason for it being disabled. This rearranges the rollout check so it’s performed as the last one.

##### Fix (flaky) test
Because the test subject was given a real randomizer, it produced mixed results - in 85% of chances leaning towards an invalid value. Passing mocked randomizer makes it stable.

**Steps to test this PR:**
1. Ensure CI is green

### DIFF
--- a/Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
+++ b/Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
@@ -195,27 +195,24 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
         let subfeatures = subfeatures(for: subfeature.parent)
         let subfeatureData = subfeatures[subfeature.rawValue]
 
-        // Handle Rollouts
-        if let rollout = subfeatureData?.rollout {
-            if !isRolloutEnabled(subfeature: subfeature, rolloutSteps: rollout.steps, randomizer: randomizer) {
-                return .disabled(.stillInRollout)
-            }
-        }
-
         let satisfiesMinVersion = satisfiesMinVersion(subfeatureData?.minSupportedVersion, versionProvider: versionProvider)
 
         switch subfeatureData?.state {
         case PrivacyConfigurationData.State.enabled:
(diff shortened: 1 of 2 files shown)


Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -195,27 +195,24 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
         let satisfiesMinVersion = satisfiesMinVersion(subfeatureData?.minSupportedVersion, versionProvider: versionProvider)
 
         switch subfeatureData?.state {
         case PrivacyConfigurationData.State.enabled:
             guard satisfiesMinVersion else { return .disabled(.appVersionNotSupported) }
-
-            return .enabled
         case PrivacyConfigurationData.State.internal:
             guard internalUserDecider.isInternalUser else { return .disabled(.limitedToInternalUsers) }
             guard satisfiesMinVersion else { return .disabled(.appVersionNotSupported) }
```
bwaresiak, line 205: We probably could extract this from 202 and 205 lines to 208 to simplify logic flow

bwaresiak (approved):
LGTM! Good job!

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
dus7: I was considering this, but it means for internal user we would enable subfeature regardless of supported version and moving these two checks before state verification would result in reporting not supported version before disabled (not sure if this is a concern). I’d prefer to keep it as it is TBH.

PR #664, opened 2024-02-15
Improve VPN rekeying reliability

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/414235014887631/1206607513978260/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2478
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2207

**Description**:

This PR fixes three issues:

1. The rekeying implementation was deleting the old key, and _then_ registering a new one - however, if the registration call failed, then it got left in a confused state with a connection that didn't work
2. The connection tester is what is responsible for triggering rekeying and the DAU pixel, but it was accidentally setting `isRunning` to false even when it was running just fine
3. The DAU event was sent only when the connection was idle, which isn't necessary

**Steps to test this PR**:
1. Make sure client PRs run and connect to NetP, try switching servers etc., generally just smoke test NetP
2. Check out this branch in one of the clients and change the `validityInterval ` in `NetworkProtectionKeyStore` to something shorter, like 30 seconds
3. Edit the `updateKeyPairExpirationDate` function to also set a date of only 30 seconds in the future, so that you can test more frequent rekeying
4. Run NetP, then watch the Console logs and check that rekeying is succeeding frequently (make sure you stop other network traffic so that the network is idle and rekeying can complete)

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
@@ -134,6 +134,7 @@ final class NetworkProtectionConnectionTester {
     func stop() async {
         os_log("🔴 Stopping connection tester", log: log)
         await stopScheduledTimer()
+        isRunning = false
     }
 
     // MARK: - Obtaining the interface
@@ -197,12 +198,10 @@ final class NetworkProtectionConnectionTester {
             self?.timer = nil
         }
 
-        isRunning = true
         timer.resume()
     }
 
     private func stopScheduledTimer() async {
-        isRunning = false
         cancelTimerImmediately()
     }
 
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
@@ -197,6 +197,7 @@ final class NetworkProtectionConnectionTester {
             self?.timer = nil
         }
 
+        isRunning = true
         timer.resume()
     }
 
@@ -242,7 +243,7 @@ final class NetworkProtectionConnectionTester {
         let onlyVPNIsDown = simulateFailure || (!vpnIsConnected && localIsConnected)
         simulateFailure = false
 
-        // After completing the conection tests we check if the tester is still supposed to be running
+        // After completing the connection tests we check if the tester is still supposed to be running
         // to avoid giving results when it should not be running.
         guard isRunning else {
             os_log("Tester skipped returning results as it was stopped while running the tests", log: log, type: .info)
--- a/Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
+++ b/Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
@@ -111,7 +111,10 @@ public final class NetworkProtectionKeychainKeyStore: NetworkProtectionKeyStore 
     }
 
     public func newKeyPair() -> KeyPair {
-        return newCurrentKeyPair()
+        let newPrivateKey = PrivateKey()
+        let newExpirationDate = Date().addingTimeInterval(validityInterval)
+
+        return KeyPair(privateKey: newPrivateKey, expirationDate: newExpirationDate)
     }
 
     private var validityInterval = Defaults.validityInterval
--- a/Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
+++ b/Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
@@ -23,7 +23,7 @@ public protocol NetworkProtectionKeyStore {
 
     /// Obtain the current `KeyPair`.
     ///
-    func currentKeyPair() -> KeyPair
+    func currentKeyPair() -> KeyPair?
 
     /// Create a new `KeyPair`.
     ///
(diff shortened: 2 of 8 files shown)


Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -197,6 +197,7 @@ final class NetworkProtectionConnectionTester {
             self?.timer = nil
         }
 
+        isRunning = true
```
samsymons, line 200: This was causing the connection tester to never call its callback block, which is what triggers rekeying and the DAU pixel. This var is set to true when `start` is first called, but then after that the class calls the `stop` function to kill any old timers that might be occurring, which then sets `isRunning` back to false.

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -203,29 +219,18 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
...
-            // but since the structure we're currently using makes the expiration date optional we need to have it.
-            // We should consider changing our server structure to not allow a missing expiration date here.
-            if let serverExpirationDate = selectedServer.expirationDate,
-               keyPair.expirationDate > serverExpirationDate {
-
-                keyPair = keyStore.updateCurrentKeyPair(newExpirationDate: serverExpirationDate)
-            }
-
-            return (selectedServer, keyPair)
+            return (selectedServer, selectedServer.expirationDate)
```
samsymons, line 226: The key pair is handled outside of this function now and passed into this func as an argument, so instead of being able to return a new one we just return a new expiration date, if one exists.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -167,11 +167,9 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
         os_log("Rekeying...", log: .networkProtectionKeyManagement)
 
-        providerEvents.fire(.rekeyCompleted)
```
samsymons, line 170: I moved this to after the `await`, to make sure that it's accurate.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -203,13 +201,13 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
         Task {
             await updateBandwidthAnalyzer()
 
+            // This provides a more frequent active user pixel check
+            providerEvents.fire(.userBecameActive)
```
samsymons, line 205: There was no reason to do this only when the connection was idle, so instead we'll do it no matter what.

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
```diff
@@ -25,19 +25,26 @@ public protocol NetworkProtectionKeyStore {
...
     ///
     /// - Parameters:
-    ///     - newExpirationDate: the new expiration date for the keypair
+    ///     - newExpirationDate: the new expiration date for the KeyPair
     ///
-    /// - Returns: a new keypair with the specified updates
+    /// - Returns: a new KeyPair with the specified updates
     ///
-    func updateCurrentKeyPair(newExpirationDate: Date) -> KeyPair
+    func updateKeyPairExpirationDate(_ newDate: Date) -> KeyPair
```
samsymons, line 47: Renamed to make sure this is easy to distinguish from the new `updateKeyPair` function.

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
```diff
@@ -103,6 +110,10 @@ public final class NetworkProtectionKeychainKeyStore: NetworkProtectionKeyStore
         return KeyPair(privateKey: currentPrivateKey, expirationDate: currentExpirationDate)
     }
 
+    public func newKeyPair() -> KeyPair {
+        return newCurrentKeyPair()
```
samsymons, line 114: This is incorrect as it'll update the key pair internally - instead we need to generate a new one but _not_store it anywhere until rekeying is done. I'll fix this.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
diegoreymendez: I saw that message in tests yesterday.

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
samsymons: Fixed now

diegoreymendez: Just curious - is this in-progress still?

samsymons: Nope, all done now - everything in all of these PRs is final, unless you find a bug. 😄

diegoreymendez:
@samsymons  - Added some suggestions here: https://github.com/duckduckgo/BrowserServicesKit/pull/666

diegoreymendez (approved):

PR #666, opened 2024-02-16
Suggestions for rekeying fixes.

diegoreymendez:
This PR just provides some suggested changes  as part of the code review of https://github.com/duckduckgo/BrowserServicesKit/pull/664

### DIFF
--- a/Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
+++ b/Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
@@ -23,7 +23,7 @@ public protocol NetworkProtectionKeyStore {
 
     /// Obtain the current `KeyPair`.
     ///
-    func currentKeyPair() -> KeyPair
+    func currentKeyPair() -> KeyPair?
 
     /// Create a new `KeyPair`.
     ///
@@ -36,15 +36,6 @@ public protocol NetworkProtectionKeyStore {
     /// Updates the existing KeyPair.
     ///
     func updateKeyPair(_ newKeyPair: KeyPair)
-
-    /// Updates the current `KeyPair` to have the specified expiration date
-    ///
-    /// - Parameters:
-    ///     - newExpirationDate: the new expiration date for the KeyPair
-    ///
-    /// - Returns: a new KeyPair with the specified updates
-    ///
-    func updateKeyPairExpirationDate(_ newDate: Date) -> KeyPair
 
     /// Resets the current `KeyPair` so a new one will be generated when requested.
     ///
@@ -61,7 +52,7 @@ public final class NetworkProtectionKeychainKeyStore: NetworkProtectionKeyStore 
     private struct Defaults {
         static let label = "DuckDuckGo Network Protection Private Key"
         static let service = "\(Bundle.main.bundleIdentifier!).privateKey"
-        static let validityInterval = TimeInterval.day
+        static let validityInterval = TimeInterval(20) //TimeInterval.day
     }
 
     private enum UserDefaultKeys {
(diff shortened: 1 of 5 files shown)


Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
```diff
@@ -23,7 +23,7 @@ public protocol NetworkProtectionKeyStore {
 
     /// Obtain the current `KeyPair`.
     ///
-    func currentKeyPair() -> KeyPair
+    func currentKeyPair() -> KeyPair?
```
diegoreymendez, line 26: This is "dumber" now so it can return `nil`.

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
```diff
@@ -37,15 +37,6 @@ public protocol NetworkProtectionKeyStore {
...
     func updateKeyPair(_ newKeyPair: KeyPair)
 
-    /// Updates the current `KeyPair` to have the specified expiration date
-    ///
-    /// - Parameters:
-    ///     - newExpirationDate: the new expiration date for the KeyPair
-    ///
-    /// - Returns: a new KeyPair with the specified updates
-    ///
-    func updateKeyPairExpirationDate(_ newDate: Date) -> KeyPair
```
diegoreymendez, line 47: This isn't great anymore, because the current keypair can be `nil`, so this method may not have anything to return.

In the actual case where this was being called, the key pair should really not be `nil`, so there's a conflict between caller and implementation.

To make things a bit cleaner, since the caller already has a valid keypair, I'm opting to just allow the caller to update the keypair data as desired (through `updateKeyPair`).  This helps us go in the direction of the store being "dumber" but also helps us maintain cleaner code.

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
```diff
@@ -82,29 +73,26 @@ public final class NetworkProtectionKeychainKeyStore: NetworkProtectionKeyStore
...
 
-    public func currentKeyPair() -> KeyPair {
+    public func currentKeyPair() -> KeyPair? {
         os_log("Querying the current key pair (publicKey: %{public}@, expirationDate: %{public}@)",
                log: .networkProtectionKeyManagement,
                String(describing: currentPublicKey),
                String(describing: currentExpirationDate))
 
         guard let currentPrivateKey = currentPrivateKey else {
-            let keyPair = newCurrentKeyPair()
```
diegoreymendez, line 92: Won't create a new keypair anymore.

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
```diff
@@ -138,13 +126,13 @@ public final class NetworkProtectionKeychainKeyStore: NetworkProtectionKeyStore
     }
 
     public func updateKeyPair(_ newKeyPair: KeyPair) {
-        self.currentPrivateKey = newKeyPair.privateKey
-        self.currentExpirationDate = newKeyPair.expirationDate
-    }
+        if currentPrivateKey != newKeyPair.privateKey {
```
diegoreymendez, line 129: Make sure we don't try to overwritte things that haven't changed.

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -132,18 +132,20 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
         if regenerateKey {
             keyPair = keyStore.newKeyPair()
         } else {
-            keyPair = keyStore.currentKeyPair()
+            keyPair = keyStore.currentKeyPair() ?? keyStore.newKeyPair()
```
diegoreymendez, line 135: In this specific call, we do want a new keypair if none exists.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -148,7 +148,11 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
     }
 
     private var isKeyExpired: Bool {
-        keyStore.currentKeyPair().expirationDate <= Date()
```
diegoreymendez, line 151: This was creating a new keypair, but not registering it with the backend.

samsymons: 🤯 Amazing catch

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeyStore.swift
samsymons: Do you think we should update the expiration date anyway, to be safe? In the case that a new key gets set here, which has a different expiration date from what this class knows about.

samsymons (approved):

PR #693, opened 2024-03-01
add history to ios

brindy:
**Required**:

Task/Issue URL: https://app.asana.com/0/0/1206696399092928/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2539
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2322

**Description**:

Adds history collection to the iOS app.

**Steps to test this PR**:

1. In the iOS app edit HistoryManager.swift and make `isHistoryFeatureEnabled()` return true
1. Browse the web a bit 
2. In the debug menu check the history section to see a list of history items
3. Use the fire button to clear the history
4. In the debug menu check the history section to confirm the entries have been removed
5. On macOS check the app builds, runs and collections history as normal (just a type name has changed there)


Sources/History/HistoryStore.swift
```diff
@@ -0,0 +1,446 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Common
+import Foundation
+import CoreData
+import Combine
+
+final public class HistoryStore: HistoryStoring {
```
brindy, line 24: This is largely copied and pasted from macOS with the encryption removed.

Tests/CommonTests/Extensions/URLExtensionTests.swift
```diff
@@ -21,6 +21,14 @@ import XCTest
 
 final class URLExtensionTests: XCTestCase {
 
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        if #available(macOS 14, *) {
+            throw XCTSkip("This test can't run on macOS 14 or higher")
+        }
```
samsymons, line 29: That's interesting, what's up with this?

Sources/History/HistoryStore.swift
```diff
@@ -0,0 +1,446 @@
...
+                    let reloadResult = self.reload(self.context)
+                    promise(reloadResult)
+                }
+            }
+        }
+    }
+
+    private func remove(_ identifiers: [UUID], context: NSManagedObjectContext) -> Result<Void, Error> {
+        // To avoid long predicate, execute multiple times
+        let chunkedIdentifiers = identifiers.chunked(into: 100)
```
samsymons, line 93: Feels like we could use array slices and iterate over windows of the array here rather than allocate a new array – pretty sure UUIDs are value types so I'd expect `chunked` is copying them needlessly.

Tests/CommonTests/Extensions/URLExtensionTests.swift
brindy: CI is still using macOS 13.  Running those tests locally fails.

Sources/History/HistoryStore.swift
brindy: This was copied from macOS and just modified to not be encrypted, sorry I should have made that clearer. I didn't want to change too much and we already discussed the possibility of refactoring this entire part of the stack to make it more efficient anyway.

samsymons (approved):

PR #703, opened 2024-03-07
Implement simple Subscription Caching mechanism using UserDefaults

afterxleep:
**Required**:

Task/Issue URL:  https://app.asana.com/0/1204099484721401/1206574474409573/f
iOS PR:  https://github.com/duckduckgo/iOS/pull/2556
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2359

**Description**:
- Implements mechanism for caching subscription info
- Caches entitlements
- Cleans cache on signout.
- Returns cached entitlements by default (if present)
- Adds a skipCache parameter to `fetchEntitlements` to force fetching from the service
- Posts a Notification only when entitlements change
- Existing API remains the same.  This does not require changes on local implementation

**Steps to test this PR**:
1.  Purchase a subscription
2. Confirm that entitlements are returned from cache after first use (Breakpoint [here](https://github.com/duckduckgo/BrowserServicesKit/pull/703/files#diff-6024bf9848884f7e6cbb4b48816ebecb139706dd2ee31fb26d08242020aff7f5R268))

---
###### Internal references:

### DIFF
--- a/Sources/Subscription/AccountManager.swift
+++ b/Sources/Subscription/AccountManager.swift
@@ -256,17 +256,17 @@ public class AccountManager: AccountManaging {
         }
 
         public func fetchEntitlements(skipCache: Bool = false) async -> Result<[Entitlement], Error> {
-            
+
             if skipCache {
                 return await fetchRemoteEntitlements()
             }
-            
+
             guard let cachedEntitlements: [Entitlement] = subscriptionCache.object(forKey: Constants.cachedEntitlementsKey) else {
                 return await fetchRemoteEntitlements()
             }
-            
+
             return .success(cachedEntitlements)
-            
+
         }
 
     public func exchangeAuthTokenToAccessToken(_ authToken: String) async -> Result<String, Error> {
--- a/Sources/Subscription/AccountManager.swift
+++ b/Sources/Subscription/AccountManager.swift
@@ -56,7 +56,7 @@ public class AccountManager: AccountManaging {
     public convenience init(subscriptionAppGroup: String) {
         let accessTokenStorage = SubscriptionTokenKeychainStorage(keychainType: .dataProtection(.named(subscriptionAppGroup)))
         self.init(accessTokenStorage: accessTokenStorage,
-                  entitlementsCache: UserDefaultsCache<[Entitlement]>(appGroup: subscriptionAppGroup, key: UserDefaultsCacheKey.subscriptionEntitlements))
+                  entitlementsCache: UserDefaultsCache<[Entitlement]>(subscriptionAppGroup: subscriptionAppGroup, key: UserDefaultsCacheKey.subscriptionEntitlements))
     }
 
     public init(storage: AccountStorage = AccountKeychainStorage(),
--- a/Sources/Subscription/AccountManager.swift
+++ b/Sources/Subscription/AccountManager.swift
@@ -22,7 +22,7 @@ import Common
 public extension Notification.Name {
     static let accountDidSignIn = Notification.Name("com.duckduckgo.subscription.AccountDidSignIn")
     static let accountDidSignOut = Notification.Name("com.duckduckgo.subscription.AccountDidSignOut")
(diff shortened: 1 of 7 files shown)


Sources/Subscription/AccountManager.swift
```diff
@@ -22,6 +22,7 @@ import Common
 public extension Notification.Name {
     static let accountDidSignIn = Notification.Name("com.duckduckgo.subscription.AccountDidSignIn")
     static let accountDidSignOut = Notification.Name("com.duckduckgo.subscription.AccountDidSignOut")
+    static let entitlementsUpdated = Notification.Name("com.duckduckgo.subscription.EntitlementsUpdated")
```
miasma13, line 25: What about `entitlementsDidChange`?

Sources/Subscription/AccountManager.swift
```diff
@@ -221,19 +230,44 @@ public class AccountManager: AccountManaging {
...
 
         switch await AuthService.validateToken(accessToken: accessToken) {
-        case .success(let response):
+
+            case .success(let response):
             let entitlements = response.account.entitlements.compactMap { Entitlement(rawValue: $0.product) }
-            return .success(entitlements)
+                if entitlements != cachedEntitlements {
+                    subscriptionCache.set(entitlements, forKey: SubscriptionCache.Component.entitlements.rawValue)
+                    NotificationCenter.default.post(name: .entitlementsUpdated, object: self, userInfo: nil)
```
miasma13, line 247: I think it would be handy to also send in the notification the updated set of entitlements.

Sources/Subscription/AccountManager.swift
```diff
@@ -221,19 +230,44 @@ public class AccountManager: AccountManaging {
...
+                return .success(entitlements)
+
+            case .failure(let error):
+                subscriptionCache.cleanup()
+                os_log(.error, log: .subscription, "[AccountManager] fetchEntitlements error: %{public}@", error.localizedDescription)
+                return .failure(error)
+            }
+        }
+
+        public func fetchEntitlements(skipCache: Bool = false) async -> Result<[Entitlement], Error> {
```
quanganhdo, line 258: Maybe I’m overthinking this, but what if instead of `skipCache` we do something similar to `NSURLRequest.CachePolicy`? E.g. supporting these scenarios based on what the API user wants:

- `reloadIgnoringLocalCacheData`: Same as `skipCache`
- `returnCacheDataElseLoad`: Use existing cache, load from remote only if there’s no cached data
- `returnCacheDataDontLoad`: Use existing cache, fails if no cached data is available

I can see the entitlement monitoring code that NetP has uses `returnCacheDataElseLoad` behavior.

Sources/Subscription/AccountManager.swift
```diff
@@ -221,19 +230,44 @@ public class AccountManager: AccountManaging {
...
             let entitlements = response.account.entitlements.compactMap { Entitlement(rawValue: $0.product) }
-            return .success(entitlements)
+                if entitlements != cachedEntitlements {
+                    subscriptionCache.set(entitlements, forKey: SubscriptionCache.Component.entitlements.rawValue)
+                    NotificationCenter.default.post(name: .entitlementsUpdated, object: self, userInfo: nil)
+                }
+                return .success(entitlements)
+
+            case .failure(let error):
+                subscriptionCache.cleanup()
```
quanganhdo, line 252: Won’t this accidentally clear the cache when there’s a connection or temporary server error?

Sources/Subscription/AccountManager.swift
quanganhdo: Second this 👍

afterxleep:
@miasma13 @quanganhdo got some updates...

1. Made the cache generic so we can not initialize for a specific type
2. Implemented Apple-like cache policy methods for elegance :bowtie: 
3. Removed cache cleanup on fetch failure
4. Added entitlements to notification object
5. Addressed small comments for vars and etc

PR #741, opened 2024-03-22
Rework the caching logic for subscription and entitlements 

miasma13:
**Required**:

Task/Issue URL: 
iOS PR: https://github.com/duckduckgo/macos-browser/pull/2485
macOS PR: https://github.com/duckduckgo/iOS/pull/2627

**Description**:
Change default staleness time interval for subscription and entitlement cache to 20min
When storing subscription in the cache use the date based on default interval or its date of renewalOrExpiry - whatever is earlier
Make the VPN code use cache and its logic for doing fetch if data is stale
On applicationDidBecomeActive refresh subscription details and entitlements
In the app rely on subscription details and entitlements based on cache logic (cached if still valid else fetch)

---
###### Internal references:

### DIFF
--- a/Sources/Subscription/AccountManager.swift
+++ b/Sources/Subscription/AccountManager.swift
@@ -56,7 +56,9 @@ public class AccountManager: AccountManaging {
     public convenience init(subscriptionAppGroup: String) {
         let accessTokenStorage = SubscriptionTokenKeychainStorage(keychainType: .dataProtection(.named(subscriptionAppGroup)))
         self.init(accessTokenStorage: accessTokenStorage,
-                  entitlementsCache: UserDefaultsCache<[Entitlement]>(subscriptionAppGroup: subscriptionAppGroup, key: UserDefaultsCacheKey.subscriptionEntitlements))
+                  entitlementsCache: UserDefaultsCache<[Entitlement]>(subscriptionAppGroup: subscriptionAppGroup, 
+                                                                      key: UserDefaultsCacheKey.subscriptionEntitlements,
+                                                                      settings: UserDefaultsCacheSettings(defaultExpirationInterval: .minutes(20))))
     }
 
     public init(storage: AccountStorage = AccountKeychainStorage(),
--- a/Sources/Subscription/AccountManager.swift
+++ b/Sources/Subscription/AccountManager.swift
@@ -53,6 +53,12 @@ public class AccountManager: AccountManaging {
         return accessToken != nil
     }
 
+    public convenience init(accessTokenStorage: SubscriptionTokenStorage) {
+        self.init(accessTokenStorage: accessTokenStorage,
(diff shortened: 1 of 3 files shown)


federicocappelli (approved):
LGTM

Sources/Subscription/AccountManager.swift
```diff
@@ -56,7 +56,9 @@ public class AccountManager: AccountManaging {
     public convenience init(subscriptionAppGroup: String) {
         let accessTokenStorage = SubscriptionTokenKeychainStorage(keychainType: .dataProtection(.named(subscriptionAppGroup)))
         self.init(accessTokenStorage: accessTokenStorage,
-                  entitlementsCache: UserDefaultsCache<[Entitlement]>(subscriptionAppGroup: subscriptionAppGroup, key: UserDefaultsCacheKey.subscriptionEntitlements))
+                  entitlementsCache: UserDefaultsCache<[Entitlement]>(subscriptionAppGroup: subscriptionAppGroup, 
+                                                                      key: UserDefaultsCacheKey.subscriptionEntitlements,
+                                                                      settings: UserDefaultsCacheSettings(defaultExpirationInterval: .minutes(20))))
```
quanganhdo, line 67: This should be a configurable param as well.

Sources/Subscription/AccountManager.swift
```diff
@@ -304,16 +306,22 @@ public class AccountManager: AccountManaging {
...
+        }
 
-        if case .success(let subscription) = await SubscriptionService.getSubscription(accessToken: token) {
+        if case .success(let subscription) = await SubscriptionService.getSubscription(accessToken: token, cachePolicy: .reloadIgnoringLocalCacheData) {
             if !subscription.isActive {
                 signOut()
             }
         }
+
+        _ = await fetchEntitlements(cachePolicy: .reloadIgnoringLocalCacheData)
```
quanganhdo, line 330: If `fetchRemoteEntitlements` returns a failure I think we should handle this. Otherwise just blocking the endpoint at router-level would make the fresh meaningless.

Sources/Subscription/Services/SubscriptionService.swift
```diff
@@ -55,7 +56,9 @@ public final class SubscriptionService: APIService {
 
         switch result {
         case .success(let subscriptionResponse):
-            subscriptionCache.set(subscriptionResponse)
+            let defaultExpiryDate = Date().addingTimeInterval(subscriptionCache.settings.defaultExpirationInterval)
```
quanganhdo, line 59: This relies on system time which can be manipulated. If device time is set to a distance future, any entitlement check that depends on the cache would just succeed. While we do use `reloadIgnoringLocalCacheData` when refreshing periodically, we don't currently handle the failure.

quanganhdo (commented):
Left some thoughts on the implementation. Not necessary blocker for merging, but they're worth pointing out IMO.

PR #748, opened 2024-03-25
Improves pixel information

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/0/1206919998699658/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2636
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2498

## Description

Adds some new errors to improve tracking of pixel errors.

## Testing

See platform-specific PRs for details.

---
###### Internal references:


Sources/NetworkProtection/Diagnostics/NetworkProtectionError.swift
```diff
@@ -93,15 +97,19 @@ public enum NetworkProtectionError: LocalizedError, CustomNSError {
...
-        case .failedToParseRegisteredServersResponse: return 106
-        case .failedToEncodeRedeemRequest: return 107
-        case .invalidInviteCode: return 108
-        case .failedToRedeemInviteCode: return 109
-        case .failedToRetrieveAuthToken: return 110
-        case .failedToParseRedeemResponse: return 111
-        case .invalidAuthToken: return 112
-        case .serverListInconsistency: return 113
+        case .noResponseFromRegisterEndpoint: return 105
+        case .unexpectedStatusFromRegisterEndpoint: return 106
```
diegoreymendez, line 101: Added these two

Sources/NetworkProtection/Diagnostics/NetworkProtectionError.swift
```diff
@@ -93,15 +97,19 @@ public enum NetworkProtectionError: LocalizedError, CustomNSError {
...
+        case .failedToFetchRegisteredServers: return 107
+        case .failedToParseRegisteredServersResponse: return 108
+        case .failedToEncodeRedeemRequest: return 109
+        case .invalidInviteCode: return 110
+        case .noResponseFromRedeemEndpoint: return 111
+        case .unexpectedStatusFromRedeemEndpoint: return 112
```
diegoreymendez, line 107: And these two

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
```diff
@@ -248,12 +256,19 @@ final class NetworkProtectionBackendClient: NetworkProtectionClient {
...
+            case 200:
+                responseData = data
+            case 401:
+                return .failure(.invalidAuthToken)
+            case 403 where isSubscriptionEnabled:
+                return .failure(.accessDenied)
+            default:
+                let domain = #function
+                let error = NSError(domain: domain, code: response.statusCode)
+                return .failure(.unexpectedStatusFromRegisterEndpoint(error))
```
diegoreymendez, line 271: Two changes here:

1. Made the code slightly more readable by adding a few newlines and separating a new case.
2. Default case no longer failing with `.failedToFetchRegisteredServers(nil)` as that was used in a few places and doesn't include status code information.  This pixel is not tracked individually so no need for a new triage.

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
```diff
@@ -248,12 +256,19 @@ final class NetworkProtectionBackendClient: NetworkProtectionClient {
         do {
             let (data, response) = try await URLSession.shared.data(for: request)
             guard let response = response as? HTTPURLResponse else {
-                return .failure(.failedToFetchRegisteredServers(nil))
+                return .failure(.noResponseFromRegisterEndpoint)
```
diegoreymendez, line 259: Changed the error to a specific one that doesn't take parameters.

This allows us to differentiate this case from the other cases below.

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
```diff
@@ -308,7 +323,7 @@ final class NetworkProtectionBackendClient: NetworkProtectionClient {
         do {
             let (data, response) = try await URLSession.shared.data(for: request)
             guard let response = response as? HTTPURLResponse else {
-                return .failure(.failedToRedeemInviteCode(nil))
+                return .failure(.noResponseFromRedeemEndpoint)
```
diegoreymendez, line 326: Now unique error to differentiate.

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
```diff
@@ -321,7 +336,9 @@ final class NetworkProtectionBackendClient: NetworkProtectionClient {
                     let decodedRedemptionResponse = try decoder.decode(AuthenticationFailureResponse.self, from: data)
                     return .failure(.failedToRetrieveAuthToken(decodedRedemptionResponse))
                 } catch {
-                    return .failure(.failedToRedeemInviteCode(nil))
+                    let domain = #function
+                    let error = NSError(domain: domain, code: response.statusCode)
+                    return .failure(.unexpectedStatusFromRedeemEndpoint(error))
```
diegoreymendez, line 341: Now a unique error with status code info.

Sources/NetworkProtection/Diagnostics/NetworkProtectionError.swift
samsymons: I notice that `failedToFetchRegisteredServers` has changed from code 105 to 107, is that intentional?

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
```diff
@@ -32,16 +32,16 @@ protocol NetworkProtectionClient {
 }
 
 public enum NetworkProtectionClientError: Error, NetworkProtectionErrorConvertible {
-    case failedToFetchLocationList(Error?)
+    case failedToFetchLocationList(Error)
```
samsymons, line 35: +1 to never allowing these to be optional

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -76,6 +79,28 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                 return "Simulated a tunnel error as requested"
             }
         }
+
+        var errorCode: Int {
+            switch self {
+                // Tunnel Setup Errors - 0+
+            case .startingTunnelWithoutAuthToken: return 0
```
samsymons, line 86: Do we need a separate error for both waitlist and sub users, since iOS and Mac App Store try to read from the subscription token store directly? It would be useful to know for example which token the extension tried and failed to fetch.

samsymons (approved):

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: As discussed privately, this may be non-trivial right now so I'm proceeding without addressing this.

Also the VPN menu app should no longer try to start the tunnel without an auth token.

Sources/NetworkProtection/Diagnostics/NetworkProtectionError.swift
diegoreymendez: As discussed via DM, I made this change as these errors are still settling and this info is still not really coming up.

But we should definitely avoid this in the future.

PR #752, opened 2024-03-26
autofill: send current language in privacy config (aka runtime config)

sjbarag:

**Required**:

Task/Issue URL: 
iOS PR: N/A
macOS PR: I'm going to link to this PR once I get a URL. One sec 😛 
What kind of version bump will this require?: Patch. This is backwards-compatible, and the autofill codebase falls back to 'en' when it doesn't receive a language.

**Optional**:

Tech Design URL: N/A
CC: @GioSensation 

**Description**:
The HTML autofill UI will be translated shortly, but doing so is only possible once that UI knows which language to use. Send the current locale's two-character language code as part of the privacy config (aka 'runtime config' in the autofill codebase).

This is a backwards-compatible change, since 'language' is currently an optional input to the autofill HTML pages.

**Steps to test this PR**:
1. Use autofill from https://github.com/duckduckgo/duckduckgo-autofill/pull/541
2. Use BSK from this PR you're currently having a read of
3. (Optional) Apply this patch to force a pseudolocale:
```diff
diff --git a/Sources/BrowserServicesKit/ContentScopeScript/ContentScopeUserScript.swift b/Sources/BrowserServicesKit/ContentScopeScript/ContentScopeUserScript.swift
index 34d1068..c4efb2d 100644
--- a/Sources/BrowserServicesKit/ContentScopeScript/ContentScopeUserScript.swift
+++ b/Sources/BrowserServicesKit/ContentScopeScript/ContentScopeUserScript.swift
@@ -34,7 +34,7 @@ public final class ContentScopeProperties: Encodable {
     public init(gpcEnabled: Bool, sessionKey: String, featureToggles: ContentScopeFeatureToggles) {
         self.globalPrivacyControlValue = gpcEnabled
         self.sessionKey = sessionKey
-        let locale = Locale.current.identifier
+        let locale = "xa_XA"
         let underscore = locale.firstIndex(of: "_")!
         self.language = String(locale[..<underscore])
         features = [
``` 
4. Build and launch the macOS app
5. Load any webpage with a signup form (I typically use IMDB)
6. Click on Dax
7. Confirm translation of the "Block email trackers" text under your personal Duck address:
    1. If you didn't apply the pseudolocale patch, you should still see English (the default in Autofill):
    image
    2. If you _did_ apply the pseudolocale patch, you should see text that almost looks like English if you squint, but also has some repeated l33tspeak-style symbols:
    image

**OS Testing**:

* [ ] iOS 14 — No changes, if I understand correctly.
* [ ] iOS 15 — No changes, if I understand correctly.
* [ ] iOS 16 — No changes, if I understand correctly.
* [ ] macOS 10.15 — I don't have 10.15 available
* [ ] macOS 11 — I don't have 11 available
* [ ] macOS 12 — Looking into it
* [x] macoS 14

---
###### Internal references:


sjbarag:
I've confirmed that this works, but I wouldn't be surprised if what I've written is extremely un-idiomatic. Please roast my Swift, for I've written very little of it 😃

GioSensation:
Thanks, @sjbarag!

@shakyShane Is there any precedent as to where this piece of data is passed? For example, I think the privacy dashboard can receive the locale on other platforms, where are we passing it? Is it worth aligning, also looking at future localisation work for Duck Player and such?

Sources/BrowserServicesKit/Autofill/AutofillUserScript+SourceProvider.swift
```diff
@@ -103,7 +103,7 @@ public class DefaultAutofillSourceProvider: AutofillUserScriptSourceProvider {
 
     private func buildReplacementsData() -> ProviderData? {
         guard let filteredPrivacyConfigData = filteredDataFrom(configData: privacyConfigurationManager.currentConfig,
-                                                               keepingTopLevelKeys: ["features", "unprotectedTemporary"],
+                                                               keepingTopLevelKeys: ["features", "unprotectedTemporary", "language"],
```
amddg44, line 106: `language` isn't actually a property of the privacy config, so there is no key to keep here to provide to `contentScope` 🙂   

It is however already available as a property in `userPreferences`: 
```
{
  "language": "en",
  "globalPrivacyControlValue": true,
  "sessionKey": "FCFAA3D3-6F11-4307-BBC2-B308935DC389",
  "features": {
    "autofill": {
      "settings": {
        "featureToggles": {
          "emailProtection_incontext_signup": true,
          "inlineIcon_credentials": true,
          "inputType_creditCards": true,
          "credentials_saving": true,
          "third_party_credentials_provider": true,
          "emailProtection": true,
          "inputType_identities": true,
          "password_generation": true,
          "inputType_credentials": true
        }
      }
    }
  },
  "debug": false,
  "platform": {
    "name": "macos"
  }
}
```

Sources/BrowserServicesKit/ContentScopeScript/ContentScopeUserScript.swift
```diff
@@ -27,12 +27,16 @@ public final class ContentScopeProperties: Encodable {
...
+    public let language: String
     public let platform = ContentScopePlatform()
     public let features: [String: ContentScopeFeature]
 
     public init(gpcEnabled: Bool, sessionKey: String, featureToggles: ContentScopeFeatureToggles) {
         self.globalPrivacyControlValue = gpcEnabled
         self.sessionKey = sessionKey
+        let locale = Locale.current.identifier
+        let underscore = locale.firstIndex(of: "_")!
+        self.language = String(locale[..<underscore])
```
amddg44, line 39: Replacing these lines with: 
`self.languageCode = Locale.current.languageCode ?? "en"` 
should work nicely for what you are doing here & will default to `en` if for some reason the system cannot determine the language code

shakyShane:
@GioSensation the dashboard uses an initial message, it's completely separate (at the moment) in terms of how it retrieves it's data. A future project to consolidate would be good, but we can ignore the dashboard in this conversation.

user preferences is the correct for the locale (outside of a separate message, which would also be fine IMO)

Sources/BrowserServicesKit/ContentScopeScript/ContentScopeUserScript.swift
sjbarag: Oooh good call! We've got a default of `en` on the autofill side as well, but happy to have one on both.

From what I saw `Locale.current.languageCode` was returning `en-US` (the autofill codebase expects just the two-letter language `en`). Maybe I'm just seeing things/mis-remembering things! Either way, I can do the string splitting on that side to keep it simple here 🙂

Sources/BrowserServicesKit/Autofill/AutofillUserScript+SourceProvider.swift
sjbarag: 🤔 Iiiiinteresting! Without adding `language` I swear I didn't get the language on the JS side. My memory is likely failing me. I'll check it out!

sjbarag:
This is going to be in limbo for a few weeks while I help out on another project. Closing for now, following the PR template's guidance. I'll be back :)

PR #764, opened 2024-04-05
Resolve location selection against remote locations

graeme:

**Required**:

Task/Issue URL: https://app.asana.com/0/1203137811378537/1206773442486707/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2688
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2805

**Description**:

The VPN geoswitching logic currently doesn’t handle the case that a location becomes unavailable. We should fall back to the country, if a city isn’t available, or nearest if the country isn’t available.

**Steps to test this PR**:
See platform PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1041,9 +993,6 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             }
 
             settings.selectedServer = .endpoint(serverName)
-            if case .connected = connectionStatus {
-                try? await updateTunnelConfiguration(environment: settings.selectedEnvironment, serverSelectionMethod: .preferredServer(serverName: serverName), reassert: true)
```
diegoreymendez, line 1045: This change is actually changing behaviour.  Why is it no longer necessary to update the tunnel configuration when the server selection changes?

Sources/NetworkProtection/Models/VPNServerSelectionSanitizer.swift
```diff
@@ -0,0 +1,106 @@
...
+public enum VPNServerSelectionError: Error {
+    case countryNotFound
+    case fetchingLocationsFailed(Error)
+}
+
+public protocol VPNServerSelectionSanitizing {
+    func sanitizedServerSelectionMethod() async -> NetworkProtectionServerSelectionMethod
+}
+
+public final class VPNServerSelectionSanitizer: VPNServerSelectionSanitizing {
```
diegoreymendez, line 30: A few comments here:

1. Let's add unit tests to cover this class in its entirety.
2. The name of this class sounds off to me.  This does more than just sanitize the selection - it effectively manages the entire selection logic (locations vs servers in settings).  Additionally when referencing this in the init parameter below (for `PacketTunnelProvider`) the instance is named `serverSelection`, which makes me think you agree.  Why not name this something more generic like `VPNServerSelector` or `VPNServerSelectionResolver`.
3. To further expand on point 2, I really think that to the `PacketTunnelProvider` whether this does sanitization or not is an implementation detail - what matters is that this class provides a server / location to connect to.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -919,35 +896,10 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-            }
-        case .setSelectedLocation(let selectedLocation):
-            let serverSelectionMethod: NetworkProtectionServerSelectionMethod
-
-            switch selectedLocation {
-            case .nearest:
-                serverSelectionMethod = .automatic
-            case .location(let location):
-                serverSelectionMethod = .preferredLocation(location)
-            }
```
diegoreymendez, line 946: Can you explain why this is all going away?  Admittedly I'm not looking to deeply into this and opting to ask as I want to avoid making (mistaken) assumptions.

diegoreymendez (commented):
I think direction is ok for helping us prevent issues... but this is not recovery.  Are we planning to follow up with proper recovery from failed registration / connection attempts?

PS: Just being extra cautious here since I'm the one asking for recovery - if we do add recovery, let's do a follow up PR to avoid this becoming huge / neverending etc.  I'm happy to keep things separated.

Sources/NetworkProtection/Models/VPNServerSelectionSanitizer.swift
graeme: Yeah, the naming here gave me a lot of trouble. So just decided to open it up to get opinions. Thanks for the input.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -919,35 +896,10 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-
+        case .setSelectedServer, .setSelectedLocation:
```
graeme, line 899: Note that this is now handling both server selection and location selection cases.

Sources/NetworkProtection/PacketTunnelProvider.swift
graeme: We still do. This now handles two cases: https://github.com/duckduckgo/BrowserServicesKit/pull/764/files#r1559075769

Sources/NetworkProtection/PacketTunnelProvider.swift
graeme: Yes sure. It’s essentially extracting this logic into the… class that has yet to be named ;). Let’s go with `ServerSelector`. I’ve been trying to avoid adding too much more logic to the PacketTunnelProvider, but this selection logic seemed closely related to the responsibilities of the `ServerSelector`. The extraction should allow it to all be unit tested too which I’ll do before moving this PR out of draft.

graeme:
@diegoreymendez I agree about this not fully handling recovery. This however, does add some improvement to the robustness of the location / server selection logic and I’m trying to go for incremental improvements. However, recovering from a failed `register` request feels a bit more generic and not just relating to locations…
 
…again though, however, I’m also not 100% sure if this is the right approach or whether the location fallback logic should not be pre-emptive as it currently is, but a reaction of a railed register request.

Sources/NetworkProtection/Models/VPNServerSelectionResolver.swift
```diff
@@ -0,0 +1,106 @@
...
+        case .location(let networkProtectionSelectedLocation):
+            serverSelectionMethod = .preferredLocation(networkProtectionSelectedLocation)
+        }
+
+        switch vpnSettings.selectedServer {
+        case .automatic:
+            break
+        case .endpoint(let string):
+            // Selecting a specific server will override locations setting
+            // Only available in debug
```
quanganhdo, line 100: Suggestion: Enforce this logic in the code (e.g. guarding `.setSelectedServer(.endpoint(String))` inside `ALPHA || DEBUG`)

Sources/NetworkProtection/Repositories/NetworkProtectionLocationListRepository.swift
```diff
@@ -21,9 +21,11 @@ import Common
 
 public protocol NetworkProtectionLocationListRepository {
     func fetchLocationList() async throws -> [NetworkProtectionLocation]
+    func fetchLocationListIgnoringCache() async throws -> [NetworkProtectionLocation]
```
quanganhdo, line 24: Suggestion: For readability, implement a `fetchLocationList(cachePolicy:)` method instead of having 2 separate ones.

Or, at least comment on how these 2 methods are used.
- `fetchLocationList()`: Display the location list on the location selector
- `fetchLocationListIgnoringCache()`: Resolve the location using fallback logic when generating tunnel config

quanganhdo (commented):
Good to see unit tests to cover different scenarios. Left some suggestions to improve code readability.

PR #772, opened 2024-04-10
PixelKit

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1206999268603553/f
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2656
What kind of version bump will this require?: Minor (new feature)

**Description**:

Migration of PixelKit to BSK and adoption in macOS ONLY

---
###### Internal references:


jaceklyp:
Hey, I hope I'm not too late to the party, but could we consider having library called PixelKit (I'm super okay with that) but class named just Pixel (instead of PixelKit)? so we can still use this API as Pixel.fire(pixelname...)?

federicocappelli:
@jaceklyp yes, that's a good idea, I'll modify the name before using the library from BSK

federicocappelli:
@jaceklyp (I deleted my previous comment) actually we may have a problem with that, in iOS the two ways of using pixels will need to coexist for a while, so having the same name would probably be a problem, we'll be able to rename PixelKit after the full migration is completed.

jaceklyp:
Ok, sounds perfect to me if we'd still consider it after migration!

PR #779, opened 2024-04-12
Bat.js fix (fix for installing content blocking rules multiple times)

jaceklyp:
Task/Issue URL: https://app.asana.com/0/1177771139624306/1205259925028360/f
iOS PR: todo
macOS PR: todo
CC: @mallexxx 

**Steps to test this PR**:
1. See main task

---
###### Internal references:


Sources/BrowserServicesKit/ContentScopeScript/UserContentController.swift
```diff
@@ -116,50 +140,75 @@ final public class UserContentController: WKUserContentController {
...
+        remove(ruleList)
     }
 
     @MainActor
     public func installLocalContentRuleList(_ ruleList: WKContentRuleList, identifier: String) {
-        localRuleLists[identifier] = ruleList
-        self.add(ruleList)
+        guard contentRuleLists[.local(identifier)] == nil else {
+            assertionFailure("Local content rule list already installed: \(contentRuleLists[.local(identifier)]!)")
+            return
```
bwaresiak, line 192: I think this is a problem.

I believe above implementation makes an implicit assumption: that every *different* rule lists set must have different identifier. This is not true, and it is also not something that is either expected on WebKit level.

The the scenario: 
1. Search for something -> navigate to vendor A. As a result we apply local rule list with "TemporaryAttributed" id, that contains exception for vendor A.
2. Go back to results -> navigate to vendor B. At this point we want to replace rules tailored for vendor A, with the ones for vendor B, but both rules share the same identifier - "TemporaryAttributed".

It means in case identifiers clash (one is already installed), we need to remove the old rules first, and just then install the new ones.

mallexxx: fixed

Sources/Common/Logging.swift
```diff
@@ -36,6 +36,7 @@ extension OSLog {
     }()
 
     public enum Categories: String, CaseIterable {
+        case contentBlocking = "Content Blocking"
```
bwaresiak, line 39: Using the same name as in macOS app repo will shows "Ambiguous use of 'contentBlocking" when I drop BSK into macOS project.

bwaresiak:
I've tested this, and I believe it solves the problem. Still to be hones I'm concerned if there is anything else we might've missed. The overall side-effect driven nature of this code makes it quite tricky to follow or to foresee the possible issues - like the one we are solving here. 

I don't think we have a good unit test suite for this, would be good to add something that:
 - Reproduces the problems we are solving here (not unloading old rule list, lack of access to up to date value in the code that listens for changes).
 - Tests the API of UserContentController.

PR #780, opened 2024-04-15
fix crash in DistributedNavigationDelegate

mallexxx:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1177771139624306/1206920871805921/f
iOS PR: not affected
macOS PR: 

**Description**:
- https://errors.duckduckgo.com/organizations/ddg/issues/17228/?query=is%3Aunresolved&referrer=issue-stream&stream_index=3
- I could not reproduce the crash but refactored the place where it happened
- fixed the wrong Navigation object chosen for quick consequent navigations (like multiple downloads)

**Steps to test this PR**:
1. the navigations handled by the modified method can be triggered by opening a `<a target="_blank" href="about:blank">link</a>`, running `window.open()` from js console and allowing the popup. Try different navigations opening a new window.
2. click "download" many times on some website, confirm no assertion is fired and correct Navigation object is used

### DIFF
--- a/Sources/Navigation/DistributedNavigationDelegate.swift
+++ b/Sources/Navigation/DistributedNavigationDelegate.swift
@@ -559,11 +559,10 @@ extension DistributedNavigationDelegate: WKNavigationDelegate {
 
             let finishedNavigationAction: NavigationAction? = {
                 guard let navigation = wkNavigation?.navigation else { return nil }
-                if navigation.state == .finished, navigation.navigationActions.isEmpty == false {
+                if navigation.state == .finished || navigation.state.isFailed, navigation.navigationActions.isEmpty == false {
                     // about: scheme navigation for `<a target='_blank'>` duplicates didStart/didCommit/didFinish events with the same WKNavigation
                     return navigation.navigationAction
                 } else {
-                    assert(webView.url?.isEmpty ?? true, "please report which steps you took to reach this state")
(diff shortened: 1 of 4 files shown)


Sources/Navigation/DistributedNavigationDelegate.swift
```diff
@@ -539,32 +539,51 @@ extension DistributedNavigationDelegate: WKNavigationDelegate {
...
+        var navigation: Navigation
+        if let approvedNavigation = wkNavigation?.navigation,
+           approvedNavigation.state == .approved, approvedNavigation.navigationActions.isEmpty == false {
+            // rely on the associated Navigation that is in the correct state
+            navigation = approvedNavigation
+
+        } else if let expectedNavigation = navigationExpectedToStart,
+                  wkNavigation != nil
+                    || expectedNavigation.navigationAction.navigationType == .sessionRestoration
+                    || expectedNavigation.navigationAction.url.scheme.map(URL.NavigationalScheme.init) == .about {
```
tomasstrba, line 551: Just a suggestion to make the code more readable. If these conditions are defined before the `if` as Bool values, their names can describe the purpose and make it easy to process for readers.

`let isUrlSchemeAbout = ...`

`if isUrlSchemeAbout || is...`

tomasstrba (approved):
LGTM! 👍 

I tested the `window.open()` call and couple of downloads.

PR #785, opened 2024-04-17
BSK: Bundle-Specfic Autofill Secure Vault Keychain Items

aataraxiaa:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1206924205159008/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2759
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2652

**Optional**:
Tech Design URL: https://app.asana.com/0/481882893211075/1206942775357851/f

**Description**:
Our two public app variants  - App Store and DMG -  share some of key pieces of Autofill-related data which are used to access their Autofill storage (Actually all variants share this data, but only public variants impact users). Specifically, the following items are shared between our App Store and DMG app variants:

* L1 Key
* L2 Ley
* L2 Key Password

This results in the following problematic scenario:

* User directly downloads and installs from our site
* On first launch of the DMG version, Autofill-related encryption data (L1 key, L2 key, password) is generated
* At this point, the user may or may not utilize our browser for Autofill data storage. Either choice does not avoid the problem
* User downloads our app from the Mac App Store
* On first launch of the App Store version, Autofill-related encryption data (L1 key, L2 key, password) is not generated, as they already exist.
* Instead the user is prompted multiple times for their password to allow the app to access the L1 key which was generated and stored as part of the initial DMG launch
* Additionally, when the app requires access to the L2 key and L2 key password (e.g launching with Sync enabled, or with Autofill data saved), they are again prompted for their password to access these Keychain items.

This scenario will also occur if the user first downloads the App Store version and then switches to the DMG. 

To address this scenario, this work implements bundle-specific Keychain storage for our L1 keys, L2 keys and L2 encryption passwords. 

We solve two specific problems:

1. Separation of Shared Keychain Data into Bundle-Specific Keychain Data
2. Migration of Existing Shared Keychain Data to Bundle-Specific Keychain Data

**Steps to test this PR**:
1. Complete all macOS tests described [here](https://github.com/duckduckgo/macos-browser/pull/2652)
2. Complete all iOS tests described [here](https://github.com/duckduckgo/iOS/pull/2759)

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

—
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/SecureVault/AutofillDatabaseProvider.swift
+++ b/Sources/BrowserServicesKit/SecureVault/AutofillDatabaseProvider.swift
@@ -809,7 +809,7 @@ extension DefaultAutofillDatabaseProvider {
         }
 
         let cryptoProvider: SecureStorageCryptoProvider = AutofillSecureVaultFactory.makeCryptoProvider()
-        let keyStoreProvider: SecureStorageKeyStoreProvider = AutofillSecureVaultFactory.makeKeyStoreProvider()
+        let keyStoreProvider: SecureStorageKeyStoreProvider = AutofillSecureVaultFactory.makeKeyStoreProvider(nil)
 
         // The initial version of the credit card model stored the credit card number as L1 data. This migration
         // updates it to store the full number as L2 data, and the suffix as L1 data for use with the Autofill
@@ -1014,7 +1014,7 @@ extension DefaultAutofillDatabaseProvider {
     static private func updatePasswordHashes(database: Database) throws {
         let accountRows = try Row.fetchCursor(database, sql: "SELECT * FROM \(SecureVaultModels.WebsiteAccount.databaseTableName)")
         let cryptoProvider: SecureStorageCryptoProvider = AutofillSecureVaultFactory.makeCryptoProvider()
-        let keyStoreProvider: SecureStorageKeyStoreProvider = AutofillSecureVaultFactory.makeKeyStoreProvider()
+        let keyStoreProvider: SecureStorageKeyStoreProvider = AutofillSecureVaultFactory.makeKeyStoreProvider(nil)
         let salt = cryptoProvider.hashingSalt
 
         while let accountRow = try accountRows.next() {
--- a/Sources/BrowserServicesKit/SecureVault/AutofillKeyStoreProvider.swift
+++ b/Sources/BrowserServicesKit/SecureVault/AutofillKeyStoreProvider.swift
@@ -52,14 +52,16 @@ final class AutofillKeyStoreProvider: SecureStorageKeyStoreProvider {
         }
 
         static func entryName(from keyValue: String) -> EntryName? {
-            if keyValue == EntryName.generatedPassword.keyValue {
+            switch keyValue {
+            case EntryName.generatedPassword.keyValue:
                 return .generatedPassword
-            } else if keyValue == EntryName.l1Key.keyValue {
+            case EntryName.l1Key.keyValue:
                 return .l1Key
-            } else if keyValue == EntryName.l2Key.keyValue {
+            case EntryName.l2Key.keyValue:
                 return .l2Key
+            default:
+                return nil
             }
-            return nil
         }
     }
 
--- a/Sources/BrowserServicesKit/SecureVault/AutofillKeyStoreProvider.swift
+++ b/Sources/BrowserServicesKit/SecureVault/AutofillKeyStoreProvider.swift
@@ -36,8 +36,8 @@ final class AutofillKeyStoreProvider: SecureStorageKeyStoreProvider {
         case l1Key = "79963A16-4E3A-464C-B01A-9774B3F695F1"
         case l2Key = "A5711F4D-7AA5-4F0C-9E4F-BE553F1EA299"
 
-        // `keyValue` should be used as Keychain Account names, as app variants (e.g App Store, DMG) should have separate entries
-        var keyValue: String {
+        // `keychainIdentifier` should be used as Keychain Account names, as app variants (e.g App Store, DMG) should have separate entries
+        var keychainIdentifier: String {
             (Bundle.main.bundleIdentifier ?? "com.duckduckgo") + rawValue
         }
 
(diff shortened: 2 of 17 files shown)


Sources/BrowserServicesKit/SecureVault/AutofillDatabaseProvider.swift
```diff
@@ -809,7 +809,7 @@ extension DefaultAutofillDatabaseProvider {
         }
 
         let cryptoProvider: SecureStorageCryptoProvider = AutofillSecureVaultFactory.makeCryptoProvider()
-        let keyStoreProvider: SecureStorageKeyStoreProvider = AutofillSecureVaultFactory.makeKeyStoreProvider()
+        let keyStoreProvider: SecureStorageKeyStoreProvider = AutofillSecureVaultFactory.makeKeyStoreProvider(nil)
```
aataraxiaa, line 812: Note: This call to `AutofillSecureVaultFactory.makeKeyStoreProvider` is part of our existing database migration logic. This is statically defined code, which does not have access to the `SecureVaultReporter` singleton in the client apps. 

This means that any Keystore-data migration that is initiated by this database migration will not fire associated pixels. However, as we are not planning any Autofill Database migrations, and none have happened recently, we believe the number of Keystore-data migrations initiated this way will be minimal.

Additionally, error reporting will still function. Pixels are for monitoring purposes to observe the number of migrations decreasing over time.

Sources/BrowserServicesKit/SecureVault/AutofillKeyStoreProvider.swift
```diff
@@ -77,28 +158,22 @@ final class AutofillKeyStoreProvider: SecureStorageKeyStoreProvider {
...
+    /// Migrates an entry to new bundle-specific Keychain storage
+    /// - Parameters:
+    ///   - entryName: Entry to migrate. It's `rawValue` is used when reading from old storage, and it's `keyValue` is used when writing to storage
+    ///   - serviceName: Service name to use when querying Keychain for the entry
+    /// - Returns: Optional data
+    private func migrateEntry(entryName: EntryName, serviceName: String) throws -> Data? {
+        guard let key = try read(named: entryName.rawValue, serviceName: serviceName) else {
             return nil
         }
+        return key
```
aataraxiaa, line 178: Note: Previously here we swallowed the error, by catching and returning `nil`. However, I believe the correct desired behavior is to propagate the error (which may relate to read or write). Once propagated, it should eventually be handled by the `SecureVaultReporter` and a pixel will be fired.

Any thoughts on this appreciated.

Sources/Common/Logging.swift
```diff
@@ -41,19 +41,21 @@ extension OSLog {
...
         case subscription = "Subscription"
         case history = "History"
+        case autofill = "Autofill"
     }
 
 #if DEBUG
     // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
     // To activate Logging Categories for DEBUG add categories here:
-    static var debugCategories: Set<Categories> = [ /* .userScripts, */ ]
+    static var debugCategories: Set<Categories> = [ .autofill ]
```
ayoy, line 50: This will be reverted before merging, right?

Tests/BrowserServicesKitTests/SecureVault/AutofillKeyStoreProviderTests.swift
```diff
@@ -0,0 +1,145 @@
...
+    private enum EntryName: String, CaseIterable {
+
+        case generatedPassword = "32A8C8DF-04AF-4C9D-A4C7-83096737A9C0"
+        case l1Key = "79963A16-4E3A-464C-B01A-9774B3F695F1"
+        case l2Key = "A5711F4D-7AA5-4F0C-9E4F-BE553F1EA299"
+
+        var keyValue: String {
+            (Bundle.main.bundleIdentifier ?? "com.duckduckgo") + rawValue
+        }
+    }
```
ayoy: You shouldn't be needing this, and just reuse types from AutofillKeyStoreProvider, right? (I'd say it's fine to mark `AutofillKeyStoreProvider.EntryName` internal vs private to expose it to the tests target).

Tests/BrowserServicesKitTests/SecureVault/AutofillKeyStoreProviderTests.swift
```diff
@@ -0,0 +1,145 @@
...
+
+            // When
+            _ = try sut.readData(named: entry.keyValue, serviceName: sut.keychainServiceName)
+
+            // Then
+            XCTAssertEqual(keychainService.itemMatchingCallCount, 3)
+        }
+    }
+
+    func testWhenReadData_AndValueNotFound_NonBundleSpecificSearchIsPerformed() throws {
```
ayoy: I think these tests names could be more precise. Currently we have:
* testWhenReadData_AndValueNotFound_AllFallbackSearchesArePerformed
* testWhenReadData_AndValueNotFound_NonBundleSpecificSearchIsPerformed
* testWhenReadData_AndValueNotFound_V1SearchIsPerformed

Which would suggest that for the same input conditions, 3 outcomes are possible. Which is not true, based on the contents of the tests.

ayoy (approved):
LGTM, great job @aataraxiaa! 👏 

Migration works great on both iOS and macOS. I'm leaving a few minor comments, it's a very solid work otherwise :)

Sources/BrowserServicesKit/SecureVault/AutofillKeyStoreProvider.swift
```diff
@@ -34,35 +36,115 @@ final class AutofillKeyStoreProvider: SecureStorageKeyStoreProvider {
...
+            case .l1Key:
+                return .l1KeyMigration
+            case .l2Key:
+                return .l2KeyMigration
+            case .generatedPassword:
+                return .l2KeyPasswordMigration
+            }
+        }
+
+        static func entryName(from keyValue: String) -> EntryName? {
```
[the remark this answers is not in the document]
aataraxiaa: Updated

Sources/Common/Logging.swift
aataraxiaa: Done

Tests/BrowserServicesKitTests/SecureVault/AutofillKeyStoreProviderTests.swift
aataraxiaa: Done

PR #786, opened 2024-04-17
VPN server failure detection recovery

graeme:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1206669340827392/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2779
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2657

**Description**:

- Use the WireGuard handshakes to determine when a connection fails by checking that the last handshake time is less than 300 seconds (just over two handshakes).
- When a failure is detected the client will call /register with server set to the name of the current (failed) server and mode set to failureRecovery. If the server is not in the "active" status on the controller a new registration will be returned otherwise the previous data will be returned (this data should be checked against the current tunnel data including allowedIPs).

**Steps to test this PR**:
1. Launch the VPN and connect 
2. Block UDP traffic by following these instructions: https://app.asana.com/0/0/1207029295107667/f
3. Open the Console.app and filter by Category: `Network Protection`. The failure recovery logs are indicated with a 🟢 so you can also look for them.
4. After around 5 minutes you will see logs indicating that a register request has been made.
5. It will error with no recovery required as there is not a real problem.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

—
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
@@ -95,7 +95,10 @@ public actor NetworkProtectionTunnelFailureMonitor {
     func stop() {
         os_log("⚫️ Stopping tunnel failure monitor", log: .networkProtectionTunnelFailureMonitorLog)
 
+        networkMonitor.cancel()
         networkMonitor.pathUpdateHandler = nil
+
+        task?.cancel() // Just making extra sure in case it's detached
         task = nil
     }
 
@@ -144,6 +147,35 @@ public actor NetworkProtectionTunnelFailureMonitor {
 }
 
 extension Network.NWPath {
+
+    /// Helper enum to identify known interfaces
+    ///
+    public enum KnownInterface: CaseIterable {
+        case utun
+        case ipsec
+        case dns
+        case unidentified
+
+        var prefix: String {
+            switch self {
+            case .utun:
+                return "utun"
+            case .ipsec:
+                return "ipsec"
+            case .dns:
+                return "dns"
+            case .unidentified:
+                return "unidentified"
+            }
+        }
+
+        static func identify(_ interface: NWInterface) -> KnownInterface {
+            allCases.first { knownInterface in
+                interface.name.hasPrefix(knownInterface.prefix)
+            } ?? .unidentified
+        }
+    }
+
     /// A description that's safe from a privacy standpoint.
     ///
     /// Ref: https://app.asana.com/0/0/1206712493935053/1206712516729780/f
@@ -157,17 +189,29 @@ extension Network.NWPath {
             description += "unsatisfiedReason: \(unsatisfiedReason), "
         }
 
-        let tunnelCount = availableInterfaces.filter { interface in
-            interface.type == .other && interface.name.contains("utun")
-        }.count
+        var dnsCount = 0
+        var ipsecCount = 0
+        var utunCount = 0
+        var unidentifiedCount = 0
 
-        let dnsCount = availableInterfaces.filter { interface in
-            interface.type == .other && interface.name.contains("dns")
-        }.count
+        availableInterfaces.map(KnownInterface.identify).forEach { knownInterface in
+            switch knownInterface {
+            case .dns:
+                dnsCount += 1
+            case .ipsec:
+                ipsecCount += 1
+            case .utun:
+                utunCount += 1
+            case .unidentified:
+                unidentifiedCount += 1
+            }
+        }
 
         description += "mainInterfaceType: \(String(describing: availableInterfaces.first?.type)), "
-        description += "tunnelInterfaceCount: \(tunnelCount), "
+        description += "utunInterfaceCount: \(utunCount), "
+        description += "ipsecInterfaceCount: \(ipsecCount), "
         description += "dnsInterfaceCount: \(dnsCount)), "
+        description += "unidentifiedInterfaceCount: \(unidentifiedCount)), "
         description += "isConstrained: \(isConstrained ? "true" : "false"), "
         description += "isExpensive: \(isExpensive ? "true" : "false")"
         description += ")"
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
@@ -52,7 +52,7 @@ public actor NetworkProtectionTunnelFailureMonitor {
         task?.isCancelled == false
     }
 
-    private weak var tunnelProvider: PacketTunnelProvider?
+    private let handshakeReporter: HandshakeReporting
 
     private let networkMonitor = NWPathMonitor()
 
@@ -61,8 +61,8 @@ public actor NetworkProtectionTunnelFailureMonitor {
 
     // MARK: - Init & deinit
 
-    init(tunnelProvider: PacketTunnelProvider) {
-        self.tunnelProvider = tunnelProvider
+    init(handshakeReporter: HandshakeReporting) {
+        self.handshakeReporter = handshakeReporter
         self.networkMonitor.start(queue: .global())
 
         os_log("[+] %{public}@", log: .networkProtectionMemoryLog, type: .debug, String(describing: self))
@@ -110,7 +110,7 @@ public actor NetworkProtectionTunnelFailureMonitor {
             return
         }
 
-        let mostRecentHandshake = await tunnelProvider?.mostRecentHandshake() ?? 0
+        let mostRecentHandshake = (try? await handshakeReporter.getMostRecentHandshake()) ?? 0
 
         guard mostRecentHandshake > 0 else {
             os_log("⚫️ Got handshake timestamp at or below 0, skipping check", log: .networkProtectionTunnelFailureMonitorLog, type: .debug)
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
@@ -157,7 +157,17 @@ extension Network.NWPath {
             description += "unsatisfiedReason: \(unsatisfiedReason), "
         }
 
-        description += "availableInterfaces: \(availableInterfaces), "
+        let tunnelCount = availableInterfaces.filter { interface in
+            interface.type == .other && interface.name.contains("utun")
+        }.count
+
+        let dnsCount = availableInterfaces.filter { interface in
+            interface.type == .other && interface.name.contains("dns")
+        }.count
+
+        description += "mainInterfaceType: \(String(describing: availableInterfaces.first?.type)), "
+        description += "tunnelInterfaceCount: \(tunnelCount), "
+        description += "dnsInterfaceCount: \(dnsCount)), "
         description += "isConstrained: \(isConstrained ? "true" : "false"), "
         description += "isExpensive: \(isExpensive ? "true" : "false")"
         description += ")"
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -31,8 +31,8 @@ public enum NetworkProtectionServerSelectionMethod: CustomDebugStringConvertible
             "avoidServer: \(serverName)"
         case .preferredLocation(let location):
             "preferredLocation: \(location)"
-        case .failureRecovery(location: let location, serverName: let serverName):
-            "failureRecovery: \(location.debugDescription), \(serverName)"
+        case .failureRecovery(serverName: let serverName):
+            "failureRecovery: \(serverName)"
         }
     }
 
@@ -40,7 +40,7 @@ public enum NetworkProtectionServerSelectionMethod: CustomDebugStringConvertible
     case preferredServer(serverName: String)
     case avoidServer(serverName: String)
     case preferredLocation(NetworkProtectionSelectedLocation)
-    case failureRecovery(location: NetworkProtectionSelectedLocation?, serverName: String)
+    case failureRecovery(serverName: String)
 }
 
 public protocol NetworkProtectionDeviceManagement {
@@ -197,8 +197,8 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
         case .preferredLocation(let location):
             serverSelection = .location(country: location.country, city: location.city)
             excludedServerName = nil
-        case .failureRecovery(location: let location, serverName: let serverName):
-            serverSelection = .recovery(server: serverName, country: location?.country, city: location?.city)
+        case .failureRecovery(serverName: let serverName):
+            serverSelection = .recovery(server: serverName)
             excludedServerName = nil
         }
 
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -59,12 +59,12 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
 
     private let isSubscriptionEnabled: Bool
 
-    public init(environment: VPNSettings.SelectedEnvironment,
+    public init(settings: VPNSettings,
                 tokenStore: NetworkProtectionTokenStore,
                 keyStore: NetworkProtectionKeyStore,
                 errorEvents: EventMapping<NetworkProtectionError>?,
                 isSubscriptionEnabled: Bool) {
-        self.init(networkClient: NetworkProtectionBackendClient(environment: environment, isSubscriptionEnabled: isSubscriptionEnabled),
+        self.init(networkClient: NetworkProtectionBackendClient(settings: settings, isSubscriptionEnabled: isSubscriptionEnabled),
                   tokenStore: tokenStore,
                   keyStore: keyStore,
                   errorEvents: errorEvents,
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -63,12 +63,12 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
 
     private let isSubscriptionEnabled: Bool
 
-    public init(settings: VPNSettings,
+    public init(environment: VPNSettings.SelectedEnvironment,
                 tokenStore: NetworkProtectionTokenStore,
                 keyStore: NetworkProtectionKeyStore,
                 errorEvents: EventMapping<NetworkProtectionError>?,
                 isSubscriptionEnabled: Bool) {
-        self.init(networkClient: NetworkProtectionBackendClient(settings: settings, isSubscriptionEnabled: isSubscriptionEnabled),
+        self.init(networkClient: NetworkProtectionBackendClient(environment: environment, isSubscriptionEnabled: isSubscriptionEnabled),
                   tokenStore: tokenStore,
                   keyStore: keyStore,
                   errorEvents: errorEvents,
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -119,7 +119,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
                                             includedRoutes: [IPAddressRange],
                                             excludedRoutes: [IPAddressRange],
                                             isKillSwitchEnabled: Bool,
-                                            regenerateKey: Bool) async throws -> (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer) {
+                                            regenerateKey: Bool) async throws -> GenerateTunnelConfigResult {
         var keyPair: KeyPair
 
         if regenerateKey {
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -49,7 +49,7 @@ public protocol NetworkProtectionDeviceManagement {
                                      includedRoutes: [IPAddressRange],
                                      excludedRoutes: [IPAddressRange],
                                      isKillSwitchEnabled: Bool,
-                                     regenerateKey: Bool) async throws -> (TunnelConfiguration, NetworkProtectionServerInfo)
+                                     regenerateKey: Bool) async throws -> (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer)
 
 }
 
@@ -118,7 +118,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
                                             includedRoutes: [IPAddressRange],
                                             excludedRoutes: [IPAddressRange],
                                             isKillSwitchEnabled: Bool,
-                                            regenerateKey: Bool) async throws -> (TunnelConfiguration, NetworkProtectionServerInfo) {
+                                            regenerateKey: Bool) async throws -> (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer) {
         var keyPair: KeyPair
 
         if regenerateKey {
@@ -156,7 +156,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
                                                         includedRoutes: includedRoutes,
                                                         excludedRoutes: excludedRoutes,
                                                         isKillSwitchEnabled: isKillSwitchEnabled)
-            return (configuration, selectedServer.serverInfo)
+            return (configuration, selectedServer)
         } catch let error as NetworkProtectionError {
             errorEvents?.fire(error)
             throw error
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -44,13 +44,13 @@ public enum NetworkProtectionServerSelectionMethod: CustomDebugStringConvertible
 }
 
 public protocol NetworkProtectionDeviceManagement {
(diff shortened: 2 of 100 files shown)


Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
```diff
@@ -52,7 +52,7 @@ public actor NetworkProtectionTunnelFailureMonitor {
         task?.isCancelled == false
     }
 
-    private weak var tunnelProvider: PacketTunnelProvider?
+    private let handshakeReporter: HandshakeReporting
```
graeme, line 55: Tried with this to remove the dependency that was there. Made it so that the Wireguard adapter is accessed directly. Alternatively could have made the `PacketTunnelProvider` implement this and kept it `weak` in a mode delegate-like fashion.

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
```diff
@@ -175,12 +191,16 @@ final class NetworkProtectionBackendClient: NetworkProtectionClient {
...
-    private let endpointURL: URL
+    private var endpointURL: URL {
+        settings.selectedEnvironment.endpointURL
+    }
+
     private let isSubscriptionEnabled: Bool
+    private let settings: VPNSettings
 
-    init(environment: VPNSettings.SelectedEnvironment = .default, isSubscriptionEnabled: Bool) {
+    init(settings: VPNSettings, isSubscriptionEnabled: Bool) {
```
graeme, line 201: This change allowed us to instantiate the client so that we don’t need to re-init the `NetworkProtectionDeviceManager` every time it's used. Should make testing a bit easier further down the line + also means we can guarantee the client always gets the most up-to-date environment setting

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -167,12 +167,18 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     // MARK: - Server Selection
 
-    public var lastSelectedServerInfo: NetworkProtectionServerInfo? {
+    @MainActor
+    private var lastSelectedServer: NetworkProtectionServer? {
```
graeme, line 171: Needed so that we can access the previous `allowedIPs`

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -756,65 +762,60 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            try await updateAdapterConfig(tunnelConfiguration: tunnelConfiguration, reassert: reassert)
+
+            providerEvents.fire(.tunnelUpdateAttempt(.success))
+        } catch {
+            providerEvents.fire(.tunnelUpdateAttempt(.failure(error)))
+        }
+    }
+
+    @MainActor
+    private func updateAdapterConfig(tunnelConfiguration: TunnelConfiguration, reassert: Bool) async throws {
```
graeme, line 779: This diff is a bit yucky, but I just pulled out the updating of the adapter config into a helper function as it’s needed for the failure recovery.

Sources/NetworkProtection/PacketTunnelProvider.swift
graeme: Also made this a `@MainActor` property, but that change somewhat snowballed and I needed to made a load of functions that access it the same. So open to other suggestions as I might be missing some details of swift concurrency that render this unnecessary.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1215,6 +1219,61 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                    return
+                }
+                await self.startReasserting()
+                await self.serverFailureMonitor.stop()
+
+                let recoveryResult: (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer)
+                do {
+                    recoveryResult = try await self.failureRecoveryHandler.attemptRecovery(
+                        to: server,
+                        location: nil, // self.settings.selectedLocation.location, TODO: This causes a bad request
```
graeme, line 1254: Asking in Asana about this.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1215,6 +1219,60 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+    lazy var failureRecoveryHandler: FailureRecoveryHandling = FailureRecoveryHandler(deviceManager: deviceManager)
+
+    private func startServerFailureMonitor() async {
+        if await serverFailureMonitor.isStarted {
+            await serverFailureMonitor.stop()
+        }
+
+        if isSubscriptionEnabled, await isEntitlementInvalid() {
+            return
+        }
```
diegoreymendez, line 1231: I just noticed we're doing this for all monitors, but I'd 100% remove this as this is not where we should be checking there's an active subscription.

The tunnel shouldn't even start without a subscription.

Once the tunnel starts (for whatever reason) the monitor should start too.

Sources/NetworkProtection/Diagnostics/NetworkProtectionTunnelFailureMonitor.swift
diegoreymendez: This looks like a good abstraction to me.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -312,7 +312,17 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+    lazy var deviceManager: NetworkProtectionDeviceManagement = NetworkProtectionDeviceManager(
+        settings: self.settings,
+        tokenStore: self.tokenStore,
+        keyStore: self.keyStore,
+        errorEvents: self.debugEvents,
+        isSubscriptionEnabled: self.isSubscriptionEnabled
+    )
+
+    private lazy var tunnelFailureMonitor = NetworkProtectionTunnelFailureMonitor(handshakeReporter: adapter)
+    private lazy var serverFailureMonitor = NetworkProtectionTunnelFailureMonitor(handshakeReporter: adapter)
```
diegoreymendez, line 324: Why do we need two failure monitors?

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,105 @@
...
+        isKillSwitchEnabled: Bool
+    ) async throws -> (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer)
+}
+
+enum FailureRecoveryError: Error {
+    case noRecoveryNecessary
+    case configGenerationError(NetworkProtectionError)
+}
+
+struct FailureRecoveryHandler: FailureRecoveryHandling {
```
diegoreymendez, line 36: Looking good, and quite testable.

Sources/NetworkProtection/Recovery/Reasserting.swift
```diff
@@ -0,0 +1,38 @@
...
+    @MainActor
+    func startReasserting() {
+        reasserting = true
+    }
+
+    @MainActor
+    func stopReasserting() {
+        reasserting = false
+    }
+}
```
diegoreymendez, line 38: It seems this isn't really used anywhere.  Can we remove this until it's needed?

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -312,7 +312,17 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-    public lazy var tunnelFailureMonitor = NetworkProtectionTunnelFailureMonitor(tunnelProvider: self)
+    lazy var deviceManager: NetworkProtectionDeviceManagement = NetworkProtectionDeviceManager(
+        settings: self.settings,
+        tokenStore: self.tokenStore,
+        keyStore: self.keyStore,
+        errorEvents: self.debugEvents,
+        isSubscriptionEnabled: self.isSubscriptionEnabled
+    )
+
+    private lazy var tunnelFailureMonitor = NetworkProtectionTunnelFailureMonitor(handshakeReporter: adapter)
```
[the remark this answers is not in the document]
diegoreymendez: I didn't understand this comment.  What was the alternative to using `NetworkProtectionTunnelFailureMonitor` and how would that alternative have affected the thresholds?

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: A few notes about this change:
- Adding a new, more generic method is fine.
- You should move the tunnel update pixels to be fired here, so the failure recovery will still fire tunnel update pixels.
- I'd consider naming that's more consistent with existing methods: `updateAdapterConfig` > `updateAdapterConfiguration`.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1215,6 +1219,60 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                        includedRoutes: self.includedRoutes ?? [],
+                        excludedRoutes: self.settings.excludedRanges,
+                        isKillSwitchEnabled: self.isKillSwitchEnabled
+                    )
+                    await self.handleFailureRecovery(tunnelConfig: recoveryResult.tunnelConfig, server: recoveryResult.server)
+                } catch {
+                    // Pixel here sent as part of https://app.asana.com/0/0/1206939413299475/f
+                    os_log("🟢 Failure recovery error: %{public}@", log: .networkProtectionServerFailureRecoveryLog, type: .error, String(reflecting: error))
+                }
+            }
```
diegoreymendez, line 1263: With the goal of trying to reduce the amount of stuff the `PacketTunnelProvider` handles, could we move this to a `ServerFailureHandler` class that's testable?

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1215,6 +1219,60 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+        }
+    }
+
+    @MainActor
+    private func handleFailureRecovery(tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer) {
+        self.lastSelectedServer = server
+        Task {
+            try await self.updateAdapterConfig(tunnelConfiguration: tunnelConfig, reassert: true)
```
diegoreymendez, line 1271: What happens here if the adapter fails to update and throws an error?

Should we add handling?  Should we fire pixels?

diegoreymendez (commented):
Mostly looking good, added some comments - some of which need to be addressed.

I'll re-review and do some extended testing once it's ready again.  Since this changes several key elements of the tunnel I want to be extra careful that everything is in place, so I'll be sinking some hours into this.

Sources/NetworkProtection/PacketTunnelProvider.swift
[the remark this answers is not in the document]
graeme: Another option would be to make the monitor configurable so that the thresholds are not hard-coded. We (Karl / Aitor / I) decided that we should ship this and see how it’s performing relative to Android then perhaps tweak the threshold if need be.

Sources/NetworkProtection/PacketTunnelProvider.swift
graeme: God that’s embarrassing actually… it made more sense when I was configuring them differently, but now I’m not. So really just need the one.

Sources/NetworkProtection/PacketTunnelProvider.swift
graeme: As that’s quite a bit change and I have two PRs stacked on this one, I’d prefer to do that in another 4th PR if that’s okay. Will add you as reviewer.

Sources/NetworkProtection/PacketTunnelProvider.swift
graeme: Yes! https://github.com/duckduckgo/BrowserServicesKit/pull/791/files

Sources/NetworkProtection/PacketTunnelProvider.swift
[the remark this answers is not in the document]
diegoreymendez: Thanks for the context.

The option of injecting is always present and doesn't need to be explained (IMO), so I think removing the comment, as you did, makes sense.

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,198 @@
...
+                    excludedRoutes: excludedRoutes,
+                    isKillSwitchEnabled: isKillSwitchEnabled
+                )
+                try await updateConfig(result)
+                eventHandler(.completed(.unhealthy))
+                await reassertingControl?.stopReasserting()
+            } catch {
+                switch error {
+                case FailureRecoveryError.noRecoveryNecessary:
+                    await reassertingControl?.stopReasserting()
```
graeme, line 107: This fixes the bug I mentioned yesterday.

Sources/NetworkProtection/Recovery/Reasserting.swift
graeme: It is now: https://github.com/duckduckgo/BrowserServicesKit/pull/786/files#r1579297467

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -755,13 +761,22 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
         do {
-            let tunnelConfiguration = try await generateTunnelConfiguration(environment: environment,
-                                                                            serverSelectionMethod: serverSelectionMethod,
+            tunnelConfiguration = try await generateTunnelConfiguration(serverSelectionMethod: serverSelectionMethod,
                                                                             includedRoutes: includedRoutes ?? [],
                                                                             excludedRoutes: settings.excludedRanges,
                                                                             regenerateKey: regenerateKey)
+        } catch {
+            providerEvents.fire(.tunnelUpdateAttempt(.failure(error)))
+            throw error
```
graeme, line 772: This means we’ll bubble up rekey failures as mentioned in https://app.asana.com/0/1203137811378537/1207149691501092/f

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
diegoreymendez: For this type of thing it's really important to ensure calls are always balanced.  If a method starts, it also should stop on every exit.  The `default` error case doesn't stop reasserting again.

The reason why this is extremely important is because otherwise you can find issues like this one, where reasserting starts but never again stops.

I see two possible solutions to this:
1. Add this right after starting the reasserting:
```swift
defer {
    await reassertingControl?.stopReasserting()
}
```
2. Since we're recovering through all the duration of the checks, why not start reasserting right before calling `incrementalPeriodicChecks` and stop reasserting right after that call?  I guess this could be an issue if something else also stops reassertions, though... so option 1 may be better.

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,198 @@
...
+                    try await action()
+                    os_log("🟢 Failure recovery success!", log: .networkProtectionServerFailureRecoveryLog, type: .info)
+                    return
+                } catch {
+                    os_log("🟢 Failure recovery failed. Retrying...", log: .networkProtectionServerFailureRecoveryLog, type: .info)
+                    try? await Task.sleep(interval: currentDelay)
+                }
+                count += 1
+                currentDelay = min((currentDelay * config.factor), config.maxDelay)
+            } while count < config.times
```
diegoreymendez, line 177: What's the expected behavior if this is exceeded?

My understanding from what I'm seeing is that:
1. We try recovery 5 times (Each time we duplicate the previous delay, but never wait more than 5 minutes)
2. Once the attempts all fail, we bail out, and the failure monitor will keep checking periodically, but we won't do any more active recovery.

Is this correct?

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,198 @@
...
+            var currentDelay = config.initialDelay
+            var count = 0
+            repeat {
+                do {
+                    try await action()
+                    os_log("🟢 Failure recovery success!", log: .networkProtectionServerFailureRecoveryLog, type: .info)
+                    return
+                } catch {
+                    os_log("🟢 Failure recovery failed. Retrying...", log: .networkProtectionServerFailureRecoveryLog, type: .info)
+                    try? await Task.sleep(interval: currentDelay)
```
diegoreymendez, line 173: This should not be ignored as its throw implies the task has been cancelled.

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,198 @@
...
+
+    private func incrementalPeriodicChecks(
+        _ config: RetryConfig,
+        action: @escaping () async throws -> Void
+    ) async {
+        let task = Task(priority: .background) {
+            var currentDelay = config.initialDelay
+            var count = 0
+            repeat {
+                do {
```
diegoreymendez, line 167: Since task cancellation is cooperative, right before the `do` I'd suggest doing:

```swift
try Task.checkCancellation()
```

Sources/NetworkProtectionTestUtils/MockNetworkProtectionDeviceManagement.swift
```diff
@@ -0,0 +1,66 @@
...
+import Foundation
+import NetworkProtection
+
+public final class MockNetworkProtectionDeviceManagement: NetworkProtectionDeviceManagement {
+    enum MockError: Error {
+        case noStubSet
+    }
+
+    // swiftlint:disable:next large_tuple
+    public var spyGenerateTunnelConfiguration: (
```
diegoreymendez, line 28: Just curious... why spy?  I'm guessing it's a concept I don't know.

diegoreymendez (changes requested):
I did some testing.  Works well so far but I'll be looking for edge cases.

Adding some comments to request changes.

The most important one is the one about stopping the reasserting.

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -59,12 +63,12 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
 
     private let isSubscriptionEnabled: Bool
 
-    public init(environment: VPNSettings.SelectedEnvironment,
+    public init(settings: VPNSettings,
```
diegoreymendez, line 66: As mentioned above this should be rolled back - we can't support updating the environment without restarting the connection.

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
quanganhdo: I also feel like replacing the `environment` param with `settings` should at least be considered outside the scope of this PR/branch. It already does a lot and refactoring this might lead to side effect.

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -115,7 +119,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
                                             includedRoutes: [IPAddressRange],
                                             excludedRoutes: [IPAddressRange],
                                             isKillSwitchEnabled: Bool,
-                                            regenerateKey: Bool) async throws -> (TunnelConfiguration, NetworkProtectionServerInfo) {
+                                            regenerateKey: Bool) async throws -> (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer) {
```
quanganhdo, line 122:
```suggestion
                                            regenerateKey: Bool) async throws -> GenerateTunnelConfigResult {
```

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -145,7 +146,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             }
 
             if case .connected = connectionStatus {
-                self.notificationsPresenter.showConnectedNotification(serverLocation: lastSelectedServerInfo?.serverLocation)
+                self.notificationsPresenter.showConnectedNotification(serverLocation: lastSelectedServer?.serverInfo.serverLocation)
```
quanganhdo, line 149:
```suggestion
                self.notificationsPresenter.showConnectedNotification(serverLocation: lastSelectedServerInfo?.serverLocation)
```

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,198 @@
...
+            do {
+                let result = try await makeRecoveryAttempt(
+                    to: lastConnectedServer,
+                    includedRoutes: includedRoutes,
+                    excludedRoutes: excludedRoutes,
+                    isKillSwitchEnabled: isKillSwitchEnabled
+                )
+                try await updateConfig(result)
+                eventHandler(.completed(.unhealthy))
+                await reassertingControl?.stopReasserting()
```
quanganhdo, line 103: In this snippet, the event is emitted before the `stopReasserting()` is invoked. In the `catch` block, it happens after the `stopReasserting()` call. The order of operations should be consistent, if only for readability.

Sources/NetworkProtection/Networking/NetworkProtectionClient.swift
graeme: I didn’t fully explain the above. The reason why it’s important to not recreate the device manager every time is that we could end up with two register requests happening at the same time which could create racy behaviour. The `NetworkProtectionDeviceManager` is an actor and it makes sense to just have one instance which will queue up these requests rather than creating new instances every time start, stop or failure recovery happens.

Rreplacing the `environment` param with `settings` was so that we could still support changing the environment without needing to restart the PacketTunnelProvider, but @diegoreymendez and I had a chat and we think it’s actually probable more sensible to require restarting as the token we’re using would become invalid.

I’m going to revert the injection of `settings` instead of `environment`, but I will still create a single instance of the `NetworkProtectionDeviceManager`. This will mean that we need to manually restart the tunnel for the env change to take effect.

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
```diff
@@ -0,0 +1,198 @@
...
+            } catch {
+                switch error {
+                case FailureRecoveryError.noRecoveryNecessary:
```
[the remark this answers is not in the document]
graeme: Yeah, fair point. It did feel a bit iffy to be honest.

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
graeme: Yeah, this is correct. I was initially throwing an error that gets bubbled up to the PacketTunnelProvider. But I wasn’t reacting to it in any way so decided it wasn’t necessary. I guess we could send another pixel in that case, but I’d rather not just now as it wasn’t in the original scope and will require another privacy triage etc.

Sources/NetworkProtectionTestUtils/MockNetworkProtectionDeviceManagement.swift
graeme: [Test jargon](https://spin.atomicobject.com/mocks-stubs-or-spies/#Spies)

Sources/NetworkProtection/Recovery/FailureRecoveryHandler.swift
graeme: Put it a different way: `reassertingControl?.startReasserting()` is the first thing that should happen, `reassertingControl?.stopReasserting()` is the last. That was my intention here. But I think that can be better expressed with a `defer`[ as Diego suggested](https://github.com/duckduckgo/BrowserServicesKit/pull/786/files#r1580000919)

graeme: Fair enough, can change that.

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -31,22 +31,26 @@ public enum NetworkProtectionServerSelectionMethod: CustomDebugStringConvertible
...
 
     case automatic
     case preferredServer(serverName: String)
     case avoidServer(serverName: String)
     case preferredLocation(NetworkProtectionSelectedLocation)
+    case failureRecovery(serverName: String)
 }
 
 public protocol NetworkProtectionDeviceManagement {
+    typealias GenerateTunnelConfigResult = (tunnelConfig: TunnelConfiguration, server: NetworkProtectionServer)
```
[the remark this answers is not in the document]
graeme: My intention was for it to be used more widely so it’s a bit more readable. Just didn’t do a great job of that.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1215,6 +1224,60 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            guard let self else {
+                return
+            }
+
+            guard case .failureDetected = result else {
+                return
+            }
+
+            Task {
+                guard let server = await self.lastSelectedServer else {
```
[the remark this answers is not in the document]
graeme: `lastSelectedServer` is a `@MainActor` var so this needs to happen in an async context.

Tests/NetworkProtectionTests/NetworkProtectionClientTests.swift
```diff
@@ -24,7 +24,7 @@ final class NetworkProtectionClientTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
-        client = NetworkProtectionBackendClient(environment: .default, isSubscriptionEnabled: false)
+        client = NetworkProtectionBackendClient(environment: .production, isSubscriptionEnabled: false)
```
diegoreymendez, line 27: This is likely a commit by mistake.

diegoreymendez (approved):

PR #755, opened 2024-03-27
Fix the convenience init not to omit the subscriptionAppGroup parameter

miasma13:
**Required**:

Task/Issue URL: https://app.asana.com/0/0/1206940718215012/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2651
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2516
What kind of version bump will this require?: Major

**Description**:
Fix convenience init for Account Manager that omits the subscriptionAppGroup param.

**Steps to test this PR**:
Test if platforms are building properly.

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #788, opened 2024-04-19
Allow Repurchase of a 3rd party subscription

afterxleep:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1204099484721401/1207112724044306/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2764/files
macOS PR: TBD

**Description**:
- Checks for the current subscription and if exired allows using the existing account to repurchase on Apple.
- Moves URLs from iOS app 
- Decouples  from subscription, to make it extensible and more generic.

**Steps to test this PR**:
1.  See iOS PR

### DIFF
--- a/Sources/Common/UserDefaultsCache.swift
+++ b/Sources/Common/UserDefaultsCache.swift
@@ -17,10 +17,13 @@
 //
 
 import Foundation
-import Common
 
 public struct UserDefaultsCacheSettings {
     public let defaultExpirationInterval: TimeInterval
+    
+    public init(defaultExpirationInterval: TimeInterval) {
+        self.defaultExpirationInterval = defaultExpirationInterval
+    }
 }
 
 public protocol UserDefaultsCacheKeyStore {
@@ -39,24 +42,18 @@ public class UserDefaultsCache<ObjectType: Codable> {
         let expires: Date
         let object: ObjectType
     }
-
-    private var subscriptionAppGroup: String?
+    
+    private var userDefaults: UserDefaults
     public private(set) var settings: UserDefaultsCacheSettings
-    private lazy var userDefaults: UserDefaults? = {
-        if let appGroup = subscriptionAppGroup {
-            return UserDefaults(suiteName: appGroup)
-        } else {
-            return UserDefaults.standard
-        }
-    }()
 
     private let key: UserDefaultsCacheKeyStore
 
-    public init(subscriptionAppGroup: String?, key: UserDefaultsCacheKeyStore,
+    public init(userDefaults: UserDefaults = UserDefaults.standard,
+                key: UserDefaultsCacheKeyStore,
                 settings: UserDefaultsCacheSettings) {
-        self.subscriptionAppGroup = subscriptionAppGroup
         self.key = key
         self.settings = settings
+        self.userDefaults = userDefaults
     }
 
     public func set(_ object: ObjectType, expires: Date? = nil) {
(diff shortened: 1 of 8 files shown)


Sources/Common/UserDefaultsCache.swift
```diff
@@ -35,24 +42,18 @@ public class UserDefaultsCache<ObjectType: Codable> {
...
-        } else {
-            return UserDefaults.standard
-        }
-    }()
 
-    private let key: UserDefaultsCacheKey
+    private let key: UserDefaultsCacheKeyStore
 
-    public init(subscriptionAppGroup: String?, key: UserDefaultsCacheKey,
+    public init(userDefaults: UserDefaults = UserDefaults.standard,
```
miasma13, line 51: Actually we got rid of default param as part of https://github.com/duckduckgo/BrowserServicesKit/pull/755
For fragile parts of code (and here deciding if we use shared or not-shared user defaults) makes more sense to be explicit in the call instead of hiding it behind default value.

Sources/Subscription/AccountManager.swift
```diff
@@ -56,15 +56,15 @@ public class AccountManager: AccountManaging {
 
     public convenience init(subscriptionAppGroup: String?, accessTokenStorage: SubscriptionTokenStorage) {
         self.init(accessTokenStorage: accessTokenStorage,
-                  entitlementsCache: UserDefaultsCache<[Entitlement]>(subscriptionAppGroup: subscriptionAppGroup,
+                  entitlementsCache: UserDefaultsCache<[Entitlement]>(userDefaults: UserDefaults.init(suiteName: subscriptionAppGroup) ?? UserDefaults.standard,
```
miasma13, line 59: Not a fan of this. WDYT about moving back to defining 2 inits one for optional `subscriptionAppGroup` and other for tests to inject UserDefaults?

Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -165,4 +170,22 @@ public final class AppStorePurchaseFlow {
...
+              let externalID = accountManager.externalID,
+              let token = accountManager.accessToken
+        else { return nil }
+        
+        let subscriptionInfo = await SubscriptionService.getSubscription(accessToken: token, cachePolicy: .reloadIgnoringLocalCacheData)
+        
+        // Only return an externalID if the subscription is expired
+        // To prevent creating multiple subscriptions in the same account
+        if case .success(let subscription) = subscriptionInfo,
+           subscription.isActive,
```
miasma13, line 185: Shouldn't it be `!subscription.isActive` ?

Sources/Subscription/AccountManager.swift
afterxleep: I have removed `subscriptionAppGroup` because this class (UserDefaultsCache) is no longer part of subscription and adding that init would imply it's coupled.   This is now an abstract class, and makes more sense to inject a fully configured UserDefaults object (or let it default to Standard)

Sources/Common/UserDefaultsCache.swift
afterxleep: I have removed subscriptionAppGroup because this class (UserDefaultsCache) is no longer part of subscription and adding that init would imply it's coupled. This is now an abstract class, and makes more sense to inject a fully configured UserDefaults object (or let it default to Standard)

miasma13 (approved):

PR #797, opened 2024-04-24
Add pixels to track VPN wake and stop attempts

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207099030609186/f

iOS PR:https://github.com/duckduckgo/iOS/pull/2785
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2694

## Description

Adds pixels to track VPN wake and stop attempts.

## Testing

See platform specific PRs for testing instructions.

---
###### Internal references:
[Pull Request Review Checklist](https://app.asana.com/0/1202500774821704/1203764234894239/f)
[Pull Request Documentation](https://app.asana.com/0/1202500774821704/1204012835277482/f)

### DIFF
--- a/Sources/Common/Concurrency/TaskExtension.swift
+++ b/Sources/Common/Concurrency/TaskExtension.swift
@@ -20,13 +20,12 @@ import Foundation
 
 public extension Task where Success == Never, Failure == Error {
 
-    static func periodic(priority: TaskPriority? = nil,
-                         delay: TimeInterval? = nil,
+    static func periodic(delay: TimeInterval? = nil,
                          interval: TimeInterval,
                          operation: @escaping @Sendable () async -> Void,
                          cancellationHandler: (@Sendable () async -> Void)? = nil) -> Task {
 
-        Task.detached(priority: priority) {
+        Task {
             do {
                 if let delay {
                     try await Task<Never, Never>.sleep(interval: delay)
@@ -42,9 +41,7 @@ public extension Task where Success == Never, Failure == Error {
                 throw error
             }
         }
-
     }
-
 }
 
 public extension Task where Success == Never, Failure == Never {
--- a/Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
+++ b/Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
@@ -51,7 +51,7 @@ final class NetworkProtectionConnectionTester {
     static let monitorQueue = DispatchQueue(label: "com.duckduckgo.NetworkProtectionConnectionTester.monitorQueue")
     static let endpoint = NWEndpoint.hostPort(host: .name("www.duckduckgo.com", nil), port: .https)
 
-    private var timer: DispatchSourceTimer?
+    private var task: Task<Never, Error>?
 
     // MARK: - Dispatch Queue
 
@@ -100,7 +100,7 @@ final class NetworkProtectionConnectionTester {
 
     deinit {
         os_log("[-] %{public}@", log: .networkProtectionMemoryLog, type: .debug, String(describing: self))
-        timer?.cancel()
+        task?.cancel()
     }
 
     // MARK: - Testing
@@ -131,9 +131,9 @@ final class NetworkProtectionConnectionTester {
         }
     }
 
-    func stop() async {
+    func stop() {
         os_log("🔴 Stopping connection tester", log: log)
-        await stopScheduledTimer()
+        stopScheduledTimer()
         isRunning = false
     }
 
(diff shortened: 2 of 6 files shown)


Sources/Common/Concurrency/TaskExtension.swift
```diff
@@ -20,13 +20,12 @@ import Foundation
...
 public extension Task where Success == Never, Failure == Error {
 
-    static func periodic(priority: TaskPriority? = nil,
-                         delay: TimeInterval? = nil,
+    static func periodic(delay: TimeInterval? = nil,
                          interval: TimeInterval,
                          operation: @escaping @Sendable () async -> Void,
                          cancellationHandler: (@Sendable () async -> Void)? = nil) -> Task {
 
-        Task.detached(priority: priority) {
```
diegoreymendez, line 29: Detached tasks need to be cancelled explicitly.  We weren't doing that anywhere.

IMHO that's likely to happen again because this `periodic` method abstracts the caller from knowing there's a detached task in here.  Since this method is only used by us right now, and it works well with a regular task, I'm just changing it to use that, instead.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -51,7 +51,7 @@ final class NetworkProtectionConnectionTester {
     static let monitorQueue = DispatchQueue(label: "com.duckduckgo.NetworkProtectionConnectionTester.monitorQueue")
     static let endpoint = NWEndpoint.hostPort(host: .name("www.duckduckgo.com", nil), port: .https)
 
-    private var timer: DispatchSourceTimer?
+    private var task: Task<Never, Error>?
```
diegoreymendez, line 54: Replaced the tester timer with a task, since my understanding is `DispatchSourceTimer` is more prone to deadlocks than a task is.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -209,10 +213,10 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             return
         }
 
-        await rekey()
+        try? await rekey()
     }
 
-    private func rekey() async {
+    private func rekey() async throws {
```
diegoreymendez, line 219: Let's rethrow so we can detect wake attempt failures due to these errors.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -664,54 +669,70 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     // MARK: - Tunnel Stop
 
-    open override func stopTunnel(with reason: NEProviderStopReason, completionHandler: @escaping () -> Void) {
```
diegoreymendez, line 667: Below this point the diff looks ugly so I'll explain the changes.

I'm changing our `stopTunnel` and `cancelTunnel` methods to be async, which allows us to track failures more easily for firing pixels.  If you see the full file at this point you'll see the new code, which IMO is much cleaner.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -989,7 +1010,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     private func handleExpireRegistrationKey(completionHandler: ((Data?) -> Void)? = nil) {
         Task {
-            await rekey()
+            try? await rekey()
             completionHandler?(nil)
```
samsymons, line 1014: I was a bit worried that we're suppressing an error value here, but I think this func is just used for the debug menu so should be fine.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -664,54 +669,70 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
     }
 
-    /// Do not cancel, directly... call this method so that the adapter and tester are stopped too.
+    // MARK: - Tunnel Stop: Support Methods
+
+    /// Do not call this directly.  Call `stopTunnel(with:)` or `cancelTunnel(with:)` instead.
+    ///
     @MainActor
-    private func cancelTunnel(with stopError: Error) async {
+    private func stopTunnel() async throws {
```
samsymons, line 712: I'd be tempted to rename this to clearly indicate it's not intended for direct use. Not a blocker though.

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: Perfect :+1:

samsymons (approved):

PR #800, opened 2024-04-25
Support subscription repurchase on Stripe

miasma13:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1207122318563030/f
iOS PR: - (only BSK version bump PR to follow)
macOS PR: 

**Optional**:

**Description**:
For Stripe subscription repurchase when authenticated to an account with an expired subscription then use its accessToken (instead of authToken).

Changes introduced in `Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift`:
- in `func prepareSubscriptionPurchase` IF user is authenticated (if access token is present) and if current subscription is expired (is in not active state) then pass access token instead of auth token for the purpose of linking purchased subscription to previous account ELSE use the old flow and create a new account and use its auth token
- in `func completeSubscriptionPurchase` only attempt auth token to access token exchange when user is not authenticated (only applies for initial subscription purchase) 

**Steps to test this PR**:
See macOS PR.

---
###### Internal references:

### DIFF
--- a/Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
+++ b/Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
@@ -66,19 +66,34 @@ public final class StripePurchaseFlow {
 
         // Clear subscription Cache
         SubscriptionService.signOut()
+        let accountManager = AccountManager(subscriptionAppGroup: subscriptionAppGroup)
 
-        var authToken: String = ""
+        var token: String = ""
 
-        switch await AuthService.createAccount(emailAccessToken: emailAccessToken) {
-        case .success(let response):
-            authToken = response.authToken
-            AccountManager(subscriptionAppGroup: subscriptionAppGroup).storeAuthToken(token: authToken)
-        case .failure:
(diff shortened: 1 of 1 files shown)


Sources/Subscription/Flows/Stripe/StripePurchaseFlow.swift
```diff
@@ -66,19 +66,34 @@ public final class StripePurchaseFlow {
...
+        } else {
+            switch await AuthService.createAccount(emailAccessToken: emailAccessToken) {
+            case .success(let response):
+                token = response.authToken
+                AccountManager(subscriptionAppGroup: subscriptionAppGroup).storeAuthToken(token: token)
+            case .failure:
+                os_log(.error, log: .subscription, "[StripePurchaseFlow] Error: accountCreationFailed")
+                return .failure(.accountCreationFailed)
+            }
+        }
```
miasma13, line 86: The whole `else` block is the previous code.

ayoy (approved):
LGTM 👏

PR #809, opened 2024-05-05
Add support for `targetPercentile` RMF rule parameter

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/0/1207234800675204/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2824
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2772

**Description**:

This PR adds support for the new `targetPercentile` option in RMF rules. This allows us to specify a percentage of users for whom the rule matches.

**Steps to test this PR**:
1. See client PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/RemoteMessaging/Model/RemoteConfigModel.swift
+++ b/Sources/RemoteMessaging/Model/RemoteConfigModel.swift
@@ -20,5 +20,15 @@ import Foundation
 
 public struct RemoteConfigModel {
     let messages: [RemoteMessageModel]
-    let rules: [Int: [MatchingAttribute]]
+    let rules: [RemoteConfigRule]
+}
+
+public struct RemoteConfigRule {
+    let id: Int
+    let targetPercentile: RemoteConfigTargetPercentile?
+    let attributes: [MatchingAttribute]
+}
+
+public struct RemoteConfigTargetPercentile {
+    let before: Float?
 }
--- a/Sources/RemoteMessaging/RemoteMessagingPercentileStoring.swift
+++ b/Sources/RemoteMessaging/RemoteMessagingPercentileStoring.swift
@@ -24,14 +24,28 @@ public protocol RemoteMessagingPercentileStoring {
 
 public class RemoteMessagingPercentileUserDefaultsStore: RemoteMessagingPercentileStoring {
 
+    enum Constants {
+        static let remoteMessagingPercentileMapping = "com.duckduckgo.app.remoteMessagingPercentileMapping"
+    }
+
     private let userDefaults: UserDefaults
 
     init(userDefaults: UserDefaults) {
         self.userDefaults = userDefaults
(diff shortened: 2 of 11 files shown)


Sources/RemoteMessaging/Model/RemoteConfigModel.swift
```diff
@@ -20,5 +20,15 @@ import Foundation
 
 public struct RemoteConfigModel {
     let messages: [RemoteMessageModel]
-    let rules: [Int: [MatchingAttribute]]
+    let rules: [RemoteConfigRule]
```
samsymons, line 23: This is technically less efficient than the old implementation, but a bit simpler. We have so few rules at one time that the difference here is negligible, but let me know if you feel otherwise.

Sources/RemoteMessaging/Model/RemoteConfigModel.swift
```diff
@@ -20,5 +20,15 @@ import Foundation
...
+}
+
+public struct RemoteConfigRule {
+    let id: Int
+    let targetPercentile: RemoteConfigTargetPercentile?
+    let attributes: [MatchingAttribute]
+}
+
+public struct RemoteConfigTargetPercentile {
+    let before: Float?
```
samsymons, line 33: This is the percentile that will be sampled _before_ we apply the attributes. The alternative here is that we do the sampling after matching attributes, but it was agreed that we'll only support the `before` case for now.

Sources/RemoteMessaging/RemoteMessagingPercentileStoring.swift
```diff
@@ -0,0 +1,51 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+
+public protocol RemoteMessagingPercentileStoring {
+    func percentile(forMessageId: String) -> Float
+}
+
+public class RemoteMessagingPercentileUserDefaultsStore: RemoteMessagingPercentileStoring {
```
samsymons, line 25: This is kept separate from the Core Data store on the iOS side, but let me know if you feel strongly.

One issue with this implementation is that percentage storage is unbounded, but I think the same is true of message storage. If I'm wrong about that, let me know and I'll look at updating this to clean up percentages.

amddg44: Correct, the message store is indeed unbounded and absolutely fine with me to keep this out of the Core Data store

amddg44 (approved):
LGTM! Have tested thoroughly over the last couple of days and this all looks really great; nice work @samsymons

PR #815, opened 2024-05-10
Subscription refactoring

federicocappelli:
Task/Issue URL: https://app.asana.com/0/72649045549333/1206805455884775/f
iOS PR: 
macOS PR: 

Tech Design URL: https://app.asana.com/0/481882893211075/1207147511614062/f
CC: @miasma13 @afterxleep @samsymons 

**Description**:

Subscription refactoring for allowing unit testing.
- DI
- Removal of all singletons
- Removal of all static functions use

**OS Testing**:
* [ ] iOS 17

---
###### Internal references:


Sources/Subscription/SubscriptionEnvironment.swift
```diff
@@ -0,0 +1,65 @@
...
+        #else
+        let platform: SubscriptionEnvironment.Platform = .appStore
+        #endif
+
+#if ALPHA || DEBUG
+        let environment: SubscriptionEnvironment.ServiceEnvironment = .staging
+#else
+        let environment: SubscriptionEnvironment.ServiceEnvironment = .production
+#endif
+        return SubscriptionEnvironment(serviceEnvironment: environment, platform: platform)
```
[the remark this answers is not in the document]
federicocappelli: fixed

Sources/Subscription/SubscriptionManager.swift
```diff
@@ -0,0 +1,132 @@
...
+        }
+        return nil
+    }
+
+    static public func save(subscriptionEnvironment: SubscriptionEnvironment, userDefaults: UserDefaults) {
+        let encoder = JSONEncoder()
+        if let encodedData = try? encoder.encode(subscriptionEnvironment) {
+            userDefaults.set(encodedData, forKey: Self.subscriptionEnvironmentStorageKey)
+        }
+    }
```
miasma13, line 90: Not sure if for now, but have you thought about removing it from the `SubscriptionManager` altogether, tbt all it requires it is to initialize it with the current value of `SubscriptionEnvironment`. How and where it is persisted or what is the current vs default logic I see more of client's responsibility, wdyt?

Sources/Subscription/SubscriptionManager.swift
```diff
@@ -0,0 +1,139 @@
...
+import Foundation
+import Common
+
+public protocol SubscriptionManaging {
+
+    var accountManager: AccountManaging { get }
+    var subscriptionService: SubscriptionService { get }
+    var authService: AuthService { get }
+    @available(macOS 12.0, iOS 15.0, *)
+    func getStorePurchaseManager() -> StorePurchaseManaging
```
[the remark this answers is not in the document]
federicocappelli: yes, this is a way to support iOS 14, `let storePurchaseManager: StorePurchaseManaging?` is optional just because StorePurchaseManagER can't be used in iOS14, but using it across the optional as optional is a huge pain, this allows the app to have a non optional StorePurchaseManagER where needed. I can change name but that seems explicative enough

Sources/Subscription/Flows/Models/SubscriptionEnvNames.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  SubscriptionEnvNames.swift
```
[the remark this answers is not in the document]
federicocappelli: No prob, renamed SubscriptionEnvironmentNames

Sources/Subscription/SubscriptionEnvironment.swift
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+    }
+
+    public enum Platform: String, Codable {
+        case appStore, stripe
+    }
+
+    public var serviceEnvironment: SubscriptionEnvironment.ServiceEnvironment
+    public var platform: SubscriptionEnvironment.Platform
```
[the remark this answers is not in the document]
federicocappelli: agree, renamed

Sources/Subscription/SubscriptionManager.swift
federicocappelli: Not sure, the UserDefaults key is shared and the logic is shared between platforms and app extensions, why would you duplicate the code? I'm ok with moving it outside SubscriptionManager into a separate entity, I was thinking of doing it as part of https://app.asana.com/0/0/1207157941206688/f

miasma13: Let's leave it for now and eventually decide in the mentioned task later.

Tests/SubscriptionTests/SubscriptionTests.swift
```diff
@@ -0,0 +1,25 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import XCTest
+@testable import Subscription
+import SubscriptionTestingUtilities
+
+final class SubscriptionTests: XCTestCase {
+
```
federicocappelli, line 24: Placeholder for step 11

Sources/SubscriptionTestingUtilities/SubscriptionManagerMock.swift
```diff
@@ -0,0 +1,59 @@
...
+    public var authService: AuthService
+    public var currentEnvironment: SubscriptionEnvironment
+    public var canPurchase: Bool
+
+    public func getStorePurchaseManager() -> StorePurchaseManaging {
+        storePurchaseManaging
+    }
+
+    public func loadInitialData() {
+
```
federicocappelli, line 35: These empty functions will be customised and refined in step 11

Sources/Subscription/SubscriptionURL.swift
```diff
@@ -0,0 +1,92 @@
...
+//
+
+import Foundation
+
+// MARK: - URLs, ex URL+Subscription
+
+public enum SubscriptionURL {
+    case baseURL
+    case purchase
+    case FAQ
```
miasma13, line 26: Please make it lowercased `FAQ` -> `faq`

Sources/Subscription/SubscriptionManager.swift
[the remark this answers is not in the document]
miasma13: Not sure if we have it as a pattern but what do you think about making it like it is done in https://github.com/duckduckgo/macos-browser/blob/15ce1f646867364c23f118c75dc3adaa1b82fbee/LocalPackages/SubscriptionUI/Sources/SubscriptionUI/DebugMenu/SubscriptionDebugMenu.swift#L41

Sources/Subscription/SubscriptionURL.swift
federicocappelli: done

Sources/Subscription/SubscriptionManager.swift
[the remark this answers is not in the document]
federicocappelli: What pattern are you referring to? A function like getXXX() that returns an instance of XXX?
The code you linked has a different purpose, serving the instance and initialising it if needed, this is only an accessor and the init is, on purpose, handled somewhere else.
What is it that you don't like? getXXX and setXXX is how getters and setters are implemented under the hood.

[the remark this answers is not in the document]
miasma13: I meant to implement it as a var and ditch the `get` prefix, as in Swift (and even before that) there was no pattern to name getters like that. In the linked example there is additional need to hide the type which has restricted availability. In general I suggest `get` less var with non-optional type and `_` prefixed storage var being optional.

Sources/Subscription/SubscriptionManager.swift
miasma13: A not for the future from usage observation. What about decoupling the `SubscriptionEnvironment` from the manager. Make `SubscriptionEnvironment` with an init that takes the `UserDefaults` and just handles the loading on init and saving on property changes.

Sources/Subscription/SubscriptionURL.swift
```diff
@@ -0,0 +1,92 @@
...
+    case addEmail
+    case manageEmail
+    case activateSuccess
+    case addEmailToSubscriptionSuccess
+    case addEmailToSubscriptionOTP
+    case manageSubscriptionsInAppStore
+    case identityTheftRestoration
+
+    // swiftlint:disable:next cyclomatic_complexity
+    public func subscriptionURL(environment: SubscriptionEnvironment.ServiceEnvironment) -> URL {
```
miasma13, line 37: Note for the future:
Going through the client PR I see this function's usage as a bit inconvenient. In my old PR I went with the route of having an URLprovider in the manager that was initialised with the current environment. Hence you just requested specific URL (by enum) from the provider. This way we could have simpler calls and instead of:
- `let url = SubscriptionURL.identityTheftRestoration.subscriptionURL(environment: Application.appDelegate.subscriptionManager.currentEnvironment.serviceEnvironment)`
there would be:
- `let url = Application.appDelegate.subscriptionManager.urlProvider.url(for: .identityTheftRestoration)`

federicocappelli: Agree, I was already working on this, replaced with a simpler `subscriptionManager.url(for: .baseURL)`

Sources/Subscription/SubscriptionManager.swift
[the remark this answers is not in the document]
federicocappelli: the iOS app contains 168 getXXX() functions, I think this is only an aesthetic preference, but I don't want to drag this out for too long.
I renamed `getStorePurchaseManager()` to `storePurchaseManager()`, I can't create a var because of the dummy class `SubscriptionManageriOS14` that needs to declare the accessor `@available(iOS 15, *)` and can't do it on a var

miasma13 (approved):

PR #819, opened 2024-05-14
Surface specific XPC & login item errors

quanganhdo:
**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1207013105069620/f
iOS PR:https://github.com/duckduckgo/iOS/pull/2872
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2773

**Optional**:

Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1. See platform-specific PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/NetworkProtection/Status/KnownFailure.swift
```diff
@@ -0,0 +1,50 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+@objc
+final public class KnownFailure: NSObject, Codable {
```
diegoreymendez, line 22: This forces all errors that are "known" to be defined in BSK which seems unnecessarily restrictive.

Why instead not just make the error adhere to `Codable` and store that in `NetworkProtectionKnownFailureStore`.

Then the VPN app can decide what errors are worth passing over IPC, but at least we don't have to add errors that are macOS specific to BSK.

diegoreymendez (commented):
Added some comments as I think there's some space for developer quality-of-life improvements.

Sources/NetworkProtection/Status/KnownFailure.swift
quanganhdo: > Why instead not just make the error adhere to Codable and store that in NetworkProtectionKnownFailureStore.

It's tricky as we need a concrete type to decode the error; and that type needs to be passed around in different ways (e.g. as part of `XPCClientInterface`, distributed notification object).

To keep it simple, I've moved the custom handling code for `TunnelError` outside of the initializer and instead made it so that an error conforming to `InternalErrorWrapping` can expose its internal error. That way we can keep specific errors outside of BSK.

Sources/NetworkProtection/ExtensionMessage/ExtensionRequest.swift
```diff
@@ -23,6 +23,7 @@ public enum VPNCommand: Codable {
     case removeSystemExtension
     case removeVPNConfiguration
     case sendTestNotification
+    case simulateKnownFailure(String, Int)
```
samsymons, line 26: Do you think it makes sense to provide an NSError here and have anyone interested in this just pull the domain and code themselves? Mainly I'd like to avoid us redefining the domain and codes ourselves and opening up the possibility of them getting out of sync with our own error types.

samsymons: Eh, after thinking about it I don't think it's worth changing, especially if this is just a debug utility.

Sources/NetworkProtection/Status/KnownFailure.swift
diegoreymendez: Understood.

That said, I do think that instead of encoding domain, code and localizedDescription, we should turn this into a `Codable` enum.  See [this comment for context](https://github.com/duckduckgo/macos-browser/pull/2773#discussion_r1617249489).

diegoreymendez (changes requested):
Added one request for a change.

diegoreymendez (approved):

PR #826, opened 2024-05-20
Add `survey` action and Privacy Pro attributes

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/1193060753475688/1207234800675206/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2879
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2798

**Description**:

This PR adds two things to RMF:

1. A new `survey` action had been added, with configurable parameters
2. New Privacy Pro attributes have been added

**Steps to test this PR**:
1. See client PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/RemoteMessaging/Mappers/RemoteMessagingSurveyActionMapping.swift
```diff
@@ -0,0 +1,35 @@
...
+    case hardwareModel = "mo"
+    case lastActiveDate = "da"
+    case osVersion = "osv"
+}
+
+public protocol RemoteMessagingSurveyActionMapping {
+
+    func add(parameters: [RemoteMessagingSurveyActionParameter], to url: URL) -> URL
+
+}
```
samsymons, line 35: This is implemented by the client. However, there would be no reason we couldn't provide a concrete implementation that the client just initializes and provides instead. I'm open to either approach.

Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
```diff
@@ -161,8 +169,23 @@ struct JsonToRemoteMessageModelMapper {
...
+                let mappedQueryParams = queryParams.compactMap { param in
+                    return RemoteMessagingSurveyActionParameter(rawValue: param)
+                }
+
+                if mappedQueryParams.count == queryParams.count, let surveyURL = URL(string: jsonAction.value) {
+                    let updatedURL = surveyActionMapper.add(parameters: mappedQueryParams, to: surveyURL)
+                    return .survey(value: updatedURL.absoluteString)
+                } else {
+                    // The message requires a parameter that isn't supported
+                    return nil
```
samsymons, line 184: @amddg44 The idea here is that if we add a new parameter and specify it in a survey, and version of the client that doesn't support the parameter will not display the message. Is this the correct way to handle that? In my testing it works as I want it to, but wanted to be sure.

amddg44: @samsymons yes, LGTM 👍

Sources/RemoteMessaging/Mappers/JsonToRemoteMessageModelMapper.swift
```diff
@@ -56,20 +57,25 @@ private enum AttributesKey: String, CaseIterable {
...
 // swiftlint:enable cyclomatic_complexity
 
 struct JsonToRemoteMessageModelMapper {
 
-    static func maps(jsonRemoteMessages: [RemoteMessageResponse.JsonRemoteMessage]) -> [RemoteMessageModel] {
+    static func maps(jsonRemoteMessages: [RemoteMessageResponse.JsonRemoteMessage],
+                     surveyActionMapper: RemoteMessagingSurveyActionMapping) -> [RemoteMessageModel] {
         var remoteMessages: [RemoteMessageModel] = []
         jsonRemoteMessages.forEach { message in
             var remoteMessage = RemoteMessageModel(id: message.id,
```
amddg44, line 74: @samsymons I stumbled across a pre-existing bug while testing - would you mind adding a fix as part of this PR? If `mapToContent` returns `nil` the message is still appended to the list of remoteMessages to display. The message won't display (since it's content is nil) but it also means that if there is a second valid message after it, that second message won't ever be shown to the user


```
            guard let content = mapToContent( content: message.content, surveyActionMapper: surveyActionMapper) else {
                return
            }
            var remoteMessage = RemoteMessageModel(id: message.id,
                                              content: content,
                                              matchingRules: message.matchingRules ?? [],
                                              exclusionRules: message.exclusionRules ?? [])
```

samsymons: Fix applied and tests updated - good spotting, thanks!

Sources/RemoteMessaging/Mappers/RemoteMessagingSurveyActionMapping.swift
samsymons: After thinking about it, I may leave this alone for now and revisit it later. I realized after the fact that we're reusing the implementation in the client for other features, so I'll avoid disrupting that if I can help it.

amddg44 (approved):
LGTM! 👏

PR #843, opened 2024-06-12
Fast egress server draining

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/1203137811378537/1207555842602667/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2945
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2858

**Description**:

This PR adds support for fast egress server draining. It works as follows:

1. Every 5 minutes, the client makes a call to the server status endpoint for the server it is connected to
2. If the response indicates that the server wants clients to transfer to another server, we will force a tunnel configuration update

**Steps to test this PR**:
1. See client PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -340,6 +340,10 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     public lazy var latencyMonitor = NetworkProtectionLatencyMonitor()
     public lazy var entitlementMonitor = NetworkProtectionEntitlementMonitor()
+    public lazy var serverStatusMonitor = NetworkProtectionServerStatusMonitor(
+        networkClient: NetworkProtectionBackendClient(isSubscriptionEnabled: true),
+        tokenStore: NetworkProtectionKeychainTokenStore()
+    )
 
     private var lastTestFailed = false
     private let bandwidthAnalyzer = NetworkProtectionConnectionBandwidthAnalyzer()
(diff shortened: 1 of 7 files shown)


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -687,10 +696,12 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                     try await self.handleAdapterStarted(startReason: startReason)
 
                     // Enable Connect on Demand when manually enabling the tunnel on iOS 17.0+.
+#if os(iOS)
```
samsymons, line 699: This probably shouldn't be in this PR but it's such a tiny change that I included it here. This is to disable the iOS connect on demand workaround on macOS entirely - right now it technically executes.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1421,6 +1458,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
         await self.tunnelFailureMonitor.stop()
         await self.latencyMonitor.stop()
         await self.entitlementMonitor.stop()
+        await self.serverStatusMonitor.stop()
```
quanganhdo, line 1461: Should we also stop this while `sleep()`?

quanganhdo (approved):
Code LGTM, approved with 2 small comments.

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: Yep, thanks for the reminder, I overlooked that we don't call this function when sleeping. Perhaps we should 🤔

PR #854, opened 2024-06-19
Improve prerequisite checks for VPN start

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/1206580121312550/1207613628544128/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2970
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2882

## Description

**Goals:**
- Validate tunnel start prerequisites and, when they aren't met:
   - Prevent pixel noise; and
   - Prevent network calls to our backend.
- Also clean up the code a bit so that we no longer use completion callbacks for the tunnel start logic.
- Add underlying error information to our wake failure pixels.

## Testing

See the platform specific PRs for testing instructions.

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -222,12 +222,18 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
             selectedServer = registeredServer
             return (selectedServer, selectedServer.expirationDate)
         case .failure(let error):
-            if isSubscriptionEnabled, case .accessDenied = error {
-                errorEvents?.fire(.vpnAccessRevoked)
-                throw NetworkProtectionError.vpnAccessRevoked
+            handle(clientError: error)
+
+            switch error {
+            case .accessDenied, .invalidAuthToken:
+                if isSubscriptionEnabled {
+                    errorEvents?.fire(.vpnAccessRevoked)
+                    throw NetworkProtectionError.vpnAccessRevoked
+                }
+            default:
+                break
             }
 
-            handle(clientError: error)
             throw error
         }
     }
--- a/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
+++ b/Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
@@ -223,17 +223,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
             return (selectedServer, selectedServer.expirationDate)
         case .failure(let error):
             handle(clientError: error)
-
-            switch error {
-            case .accessDenied, .invalidAuthToken:
-                if isSubscriptionEnabled {
-                    errorEvents?.fire(.vpnAccessRevoked)
-                    throw NetworkProtectionError.vpnAccessRevoked
-                }
-            default:
-                break
-            }
-
+            try handleAccessRevoked(error)
             throw error
         }
     }
@@ -324,5 +314,17 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
         }
 #endif
         errorEvents?.fire(clientError.networkProtectionError)
+    }
+
+    private func handleAccessRevoked(_ error: NetworkProtectionClientError) throws {
+        switch error {
+        case .accessDenied, .invalidAuthToken:
+            if isSubscriptionEnabled {
+                errorEvents?.fire(.vpnAccessRevoked)
+                throw NetworkProtectionError.vpnAccessRevoked
+            }
+        default:
+            break
+        }
     }
 }
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -1469,9 +1469,25 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     // MARK: - Connection Tester
 
-    private enum ConnectionTesterError: Error {
+    private enum ConnectionTesterError: CustomNSError {
         case couldNotRetrieveInterfaceNameFromAdapter
         case testerFailedToStart(internalError: Error)
+
+        var errorCode: Int {
+            switch self {
+            case .couldNotRetrieveInterfaceNameFromAdapter: return 0
+            case .testerFailedToStart: return 1
+            }
+        }
+
+        var errorUserInfo: [String : Any] {
+            switch self {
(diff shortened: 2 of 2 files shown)


Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -222,12 +222,18 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
             selectedServer = registeredServer
             return (selectedServer, selectedServer.expirationDate)
         case .failure(let error):
-            if isSubscriptionEnabled, case .accessDenied = error {
-                errorEvents?.fire(.vpnAccessRevoked)
-                throw NetworkProtectionError.vpnAccessRevoked
+            handle(clientError: error)
+
+            switch error {
+            case .accessDenied, .invalidAuthToken:
```
diegoreymendez, line 228: First change, we treat `.invalidAuthToken` as `.accessDenied`.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -516,6 +516,9 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
             try tokenStore.store(newAuthToken)
         case .useExisting:
+            guard try tokenStore.fetchToken() != nil else {
+                throw TunnelError.startingTunnelWithoutAuthToken
+            }
```
diegoreymendez, line 521: If we plan to use the existing auth token we might as well make sure there is one.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -564,69 +566,66 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                // manual start attempt that preceded failed, or if the subscription has
+                // expired.  In either case it should be enough to record the manual failures
+                // for these prerequisited to avoid flooding our metrics.
+                try? await Task.sleep(interval: .seconds(15))
+            } else {
+                // If the VPN was started manually without the basic prerequisited we always
+                // want to know as this should not be possible.
+                providerEvents.fire(.tunnelStartAttempt(.begin))
+                providerEvents.fire(.tunnelStartAttempt(.failure(error)))
+            }
```
diegoreymendez, line 591: This is the gist of the changes... we check prerequisites before all else.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -564,69 +566,66 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-                        try? await Task.sleep(interval: .seconds(10))
-                        handler()
-                    }
-                } else {
-                    handler()
-                }
+            providerEvents.fire(.tunnelStartAttempt(.success))
+        } catch {
+            if startupOptions.startupMethod == .automaticOnDemand {
+                // We add a 10 seconds delay when the VPN is started by
```
quanganhdo, line 611:
```suggestion
                // We add a delay when the VPN is started by
```

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -564,69 +566,66 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            throw error
+        }
 
-                    os_log("Tunnel startup error: %{public}@", type: .error, errorDescription)
-                    self?.controllerErrorStore.lastErrorMessage = errorDescription
-                    self?.connectionStatus = .disconnected
-                    self?.knownFailureStore.lastKnownFailure = KnownFailure(error)
+        do {
+            providerEvents.fire(.tunnelStartAttempt(.begin))
+            prepareToConnect(using: tunnelProviderProtocol)
```
quanganhdo, line 601: We should invoke this as soon as possible, as `PixelKit` is expected to be set up within an overriden `prepareToConnect` implementation.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -564,69 +566,66 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     // MARK: - Tunnel Start
 
-    open override func startTunnel(options: [String: NSObject]?, completionHandler: @escaping (Error?) -> Void) {
-        Task { @MainActor in
-            providerEvents.fire(.tunnelStartAttempt(.begin))
-            prepareToConnect(using: tunnelProviderProtocol)
```
quanganhdo, line 570: As `prepareToConnect` is necessary to set up PixelKit, we might've undercounted `tunnel-start-attempt-begin` 🤔

samsymons: I think we're fine, but the existing logic is a little redundant. `prepareToConnect` does indeed set up pixels, but **only** if they haven't already been set up, and thankfully we also call `setupPixels` during packet tunnel provider init. I'd guess that the set-up call within `prepareToConnect` never runs because it's always set up at this point.

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: Well, that's a very good observation.  Moved to the top.

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: I think there's merit to this observation.

The call to `loadVendorOptions` is setting up PixelKit but if can fail before it's set up.  I think this call is currently needed, and moving it to the top seems aligned with the spirit of it, so for simplicity I'll do that.

We can review this code at a second time, but moving this to the top should be fairly safe.

diegoreymendez:
Ready for another look.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1470,9 +1469,25 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+        }
+
+        var errorUserInfo: [String: Any] {
+            switch self {
+            case .couldNotRetrieveInterfaceNameFromAdapter:
+                return [:]
+            case .testerFailedToStart(let internalError):
+                return [NSUnderlyingErrorKey: internalError as NSError]
+            }
+        }
```
diegoreymendez, line 1490: We were missing wake error information from our pixels since this didn't implement `CustomNSError` so I just added that to avoid another review.

quanganhdo (approved):

PR #855, opened 2024-06-19
Custom DNS for VPN

quanganhdo:

**Required**:

Task/Issue URL: https://app.asana.com/0/1107339934162485/1204085249612942/f
iOS PR: https://github.com/duckduckgo/iOS/pull/2971
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2883

**Optional**:

Tech Design URL:
CC:

**Description**:

Adds support for custom DNS settings.

**Steps to test this PR**:

See platform-specific PRs. 

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -270,11 +291,21 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
...
         }
 
+        let dns: [DNSServer]
+        switch dnsSettings {
+        case .default:
+            dns = [DNSServer(address: server.serverInfo.internalIP)]
+        case .custom(let servers):
+            dns = servers
+                .compactMap { IPv4Address($0) }
+                .map { DNSServer(address: $0) }
```
samsymons, line 301: Since we have two values for storing DNS settings, one for indicating whether to use custom DNS and another with the DNS servers themselves, it's possible for this to break if the DNS server array is unavailable. At that point the app will reach this code and end up with an empty DNS server array. It seems very unlikely, but on macOS users can modify user default values manually via the Terminal so anything's possible.

Should we fall back to using our own DNS server if this happens?

Sources/NetworkProtection/Settings/Extensions/UserDefaults+dnsSettings.swift
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    private var dnsSettingKey: String {
+        "dnsSettingStorageValue"
+    }
+
+    private static func dnsSettingsFromStorageValue(_ value: StorableDNSSettings) -> NetworkProtectionDNSSettings {
+        guard value.usesCustomDNS else { return .default }
+        return .custom(value.dnsServers)
```
samsymons, line 39: Actually, related to the previous comment, here is where we should likely have some recovery logic in the case that `value.usesCustomDNS` is true but `value.dnsServers` is empty for some reason. (Probably it can have an assertion failure as well, to let us know locally if this ever happens.)

samsymons, line 39: Are there any guards we should add for the case where `dnsServers` is somehow empty?

samsymons (commented):
What happens in the case where the user enters an IP address which isn't a valid DNS server?

Sources/NetworkProtection/Settings/Extensions/UserDefaults+dnsSettings.swift
quanganhdo: Good catch 👍

quanganhdo:
@samsymons Updated the code so that we'll do a best-effort conversion to IPv4 address, and if it fails, fallback to default.

Sources/Common/Extensions/StringExtension.swift
```diff
@@ -110,6 +110,15 @@ public extension String {
         return false
     }
 
+    var isValidIpv4Host: Bool {
+        guard let toIPv4Host, !toIPv4Host.isEmpty else { return false }
+        return true
+    }
+
+    var toIPv4Host: String? {
+        IPv4Address(self)?.debugDescription
```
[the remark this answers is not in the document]
samsymons: Agreed that using `debugDescription` should be discouraged, there's no guarantee Apple won't change the output of this underneath us in the future. Instead, we could use `address.rawValue` to get the bytes, and build the string ourselves. Something like:

```
let address = IPv4Address("1")!
let byteArray = [UInt8](address.rawValue)
let ipAddress = byteArray.map({ String($0) }).joined(separator: ".")

print(ipAddress) // "0.0.0.1"
```

Tests/CommonTests/Extensions/StringExtensionTests.swift
```diff
@@ -88,4 +88,9 @@ final class StringExtensionTests: XCTestCase {
         XCTAssertEqual("about:blank#navlink1".url!.absoluteString.droppingHashedSuffix(), "about:blank")
     }
 
+    func testToIPv4Host() {
+        XCTAssertEqual("1.1.1.1".toIPv4Host, "1.1.1.1")
+        XCTAssertEqual("1".toIPv4Host, "0.0.0.1")
+        XCTAssertEqual("1.2".toIPv4Host, "1.0.0.2")
```
samsymons, line 94: Nice :+1:

Sources/NetworkProtection/Settings/Extensions/UserDefaults+dnsSettings.swift
```diff
@@ -0,0 +1,85 @@
...
+
+extension UserDefaults {
+    final class StorableDNSSettings: NSObject, Codable {
+        let usesCustomDNS: Bool
+        let dnsServers: [String]
+
+        init(usesCustomDNS: Bool = false, dnsServers: [String] = []) {
+            self.usesCustomDNS = usesCustomDNS
+            self.dnsServers = dnsServers
+        }
```
samsymons, line 30: Controversial suggestion: what if we make this a failable initializer, which returns nil if `usesCustomDNS` is true but `dnsServers` doesn't meet our criteria? (i.e., the array is empty, the strings aren't valid IPs, etc.)

quanganhdo: @samsymons I'm doing the data sanitization in the setter; enforcing this in StorableDNSSettings feels a level too low, if that makes sense?

https://github.com/duckduckgo/BrowserServicesKit/blob/f63347234551c90fd3676dbf9f41d4e9b2a6c0a9/Sources/NetworkProtection/Settings/Extensions/UserDefaults%2BdnsSettings.swift#L61-L73

samsymons: Overall I'd say the current solution is fine, mainly what I'm getting at is that it's possible today to create an invalid  `StorableDNSSettings` object and having it done at the init level would make it impossible. But, given the usage here is so specific, I think we can go with it. :+1:

samsymons (approved):

PR #876, opened 2024-07-03
Remote Messaging Framework for macOS

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1202913520695928/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3031
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2937

**Optional**:

CC: @amddg44 @samsymons 

**Description**:
This change adds support for RMF on macOS.
* Remaining RMF functionality previously present in iOS codebase was moved to BSK
  (most importantly RemoteMessagingProcessing protocol).
* remoteMessaging feature flag was added.
* RemoteMessagingConfigMatcherProviding protocol was added – it's implemented by client apps
  to create and return the config matcher filled with current values of matching attributes.
* Support for platform-specific user attribute matchers was added. CommonUserAttributeMatcher was added
  to handle attributes that are the supported on both iOS and macOS. Mobile and Desktop user attribute matcher
  structs were added to support platform-specific attributes (currently it's only isWidgetInstalled for iOS).
* RemoteMessagingRequest was replaced with RemoteMessagingConfigFetcher that uses Configuration Store
  and supports Etags.
* All RMF tests (and mocks) were moved out of BrowserServicesKit(Tests) into RemoteMessagingTestsUtils
  and RemoteMessagingTests (that also mandated creating EmailTestsUtils).
* RemoteMessagingStore was updated to keep a reference to the database and create contexts on demand,
  to ensure it always gets the freshest data from the persistent store.

**Steps to test this PR**:
See [iOS PR](https://github.com/duckduckgo/iOS/pull/3031) for steps to test.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/RemoteMessaging/RemoteMessagingStore.swift
```diff
@@ -47,45 +47,45 @@ public final class RemoteMessagingStore: RemoteMessagingStoring {
         public static let privateContextName = "RemoteMessaging"
     }
 
-    let context: NSManagedObjectContext
+    let database: CoreDataDatabase
```
ayoy, line 50: This change makes the functions in RemoteMessagingStore create context as needed, ensuring that they have the fresh data all the time. We use a similar approach in Sync Bookmarks provider.

Sources/RemoteMessagingTestsUtils/MockRemoteMessagePercentileStore.swift
```diff
@@ -19,12 +19,17 @@
 import Foundation
 import RemoteMessaging
 
-class MockRemoteMessagePercentileStore: RemoteMessagingPercentileStoring {
+public class MockRemoteMessagePercentileStore: RemoteMessagingPercentileStoring {
```
ayoy, line 22: Mocks moved to a dedicated mocks target (RemoteMessagingTestsUtils in this case) need to be made public so that the module could be imported without `@testable`

Package.swift
```diff
@@ -166,6 +166,12 @@ let package = Package(
                 .define("DEBUG", .when(configuration: .debug))
             ]
         ),
+        .target(
+            name: "EmailTestsUtils",
```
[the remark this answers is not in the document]
ayoy: EDIT: I renamed it as `BrowserServicesKitTestsUtils` to indicate that these are mocks of classes from `BrowserServicesKit` module. Also moved there StatisticsStore, Variant and VariantManager mocks.

Sources/RemoteMessaging/RemoteMessagingProcessing.swift
```diff
@@ -0,0 +1,101 @@
...
+        RemoteMessagingConfigProcessor(remoteMessagingConfigMatcher: configMatcher)
+    }
+
+    func fetchAndProcess(using store: RemoteMessagingStoring) async throws {
+        guard remoteMessagingAvailabilityProvider.isRemoteMessagingAvailable else {
+            os_log("Remote messaging feature flag is disabled, skipping fetching messages", log: .remoteMessaging, type: .debug)
+            return
+        }
+        do {
+            let statusResponse = try await configurationFetcher.fetchRemoteMessagingConfig()
```
jaceklyp, line 83: I think this might be some old naming, where this func used to return status code or something? But now it is config indeed, isn't it?

I would actually suggest a couple of helper func here to make this functionality here more readable, e.g:
private func fetchRemoteMessagingConfig() async throws -> RemoteMessagingConfig
private func processRemoteMessagingConfig(_ config: RemoteMessagingConfig, using store: RemoteMessagingStoring) async throws -> ProcessedResult

where first func just implements: try await configurationFetcher.fetchRemoteMessagingConfig()
second: pretty much old the rest

Sources/RemoteMessaging/RemoteMessagingProcessing.swift
```diff
@@ -0,0 +1,101 @@
...
+            os_log("Successfully fetched remote messages", log: .remoteMessaging, type: .debug)
+
+            let remoteMessagingConfigMatcher = await configMatcherProvider.refreshConfigMatcher(using: store)
+
+            let processor = configProcessor(for: remoteMessagingConfigMatcher)
+            let config = store.fetchRemoteMessagingConfig()
+
+            if let processorResult = processor.process(jsonRemoteMessagingConfig: statusResponse, currentConfig: config) {
+                store.saveProcessedResult(processorResult)
+            }
```
jaceklyp, line 94: perhaps this should be guard that throws some processing failed error? otherwise we're not getting any debug info what happened

jaceklyp (approved):
looks solid :)

Sources/RemoteMessaging/RemoteMessagingProcessing.swift
ayoy: Thanks! Yeah, I agree it's a bit messy. This function used to be called `fetchRemoteMessages`. Now I've renamed it but failed to notice that there are two `fetchRemoteMessagingConfig` functions being called here, from 2 different protocols 🤦  and they do 2 different things, first one fetches the config from the server, and the second one fetches config metadata from the local database. 😬 

Let me try to make it right :)

Sources/RemoteMessaging/RemoteMessagingProcessing.swift
ayoy: This method calls a bunch of Store private methods inside, each of which can throw an error that is handled by sending a pixel (via `EventMapping`) and printing a console log. So I think it's acceptable. I agree that it's not ideal though.

Also, this function's error is silenced at the call site (with `try?` – again probably acceptable in this specific case because it effectively silences only network and JSON parsing errors). I may propose an update to this piece of code but it has been a status quo for past 2 years so I'd rather handle it separately.

jaceklyp: don't bother, makes sense to me

jaceklyp (approved):

PR #897, opened 2024-07-17
Prevent endless rekey and tunnel update cycle

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/1207603085593419/1207806947583090/f

iOS PR: https://github.com/duckduckgo/iOS/pull/3091
macOS PR: https://github.com/duckduckgo/macos-browser/pull/2987

## Description

Prevents the following endless loop of calls `updateTunnel` > `rekey` > `updateTunnel` > `rekey`.

This adds noise to our metrics and is unnecessary.

## Testing

See platform-specific PRs for testing instructions.

**Definition of Done**:

* [ ] Does this PR satisfy our [Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f)?

---
###### Internal references:
[Pull Request Review Checklist](https://app.asana.com/0/1202500774821704/1203764234894239/f)
[Pull Request Documentation](https://app.asana.com/0/1202500774821704/1204012835277482/f)


Sources/Common/Concurrency/TaskExtension.swift
```diff
@@ -42,6 +42,29 @@ public extension Task where Success == Never, Failure == Error {
             }
         }
     }
+
+    static func periodic(delay: TimeInterval? = nil,
+                         interval: TimeInterval,
+                         operation: @escaping @Sendable () async throws -> Void,
```
diegoreymendez, line 48: A different version of this method that allows operation to throw.  Throwing within tasks cancels them, so this is rather standard.

Sources/NetworkProtection/Diagnostics/NetworkProtectionEntitlementMonitor.swift
```diff
@@ -52,7 +52,7 @@ public actor NetworkProtectionEntitlementMonitor {
 
     // MARK: - Start/Stop monitoring
 
-    public func start(entitlementCheck: @escaping () async -> Swift.Result<Bool, Error>, callback: @escaping (Result) -> Void) {
+    public func start(entitlementCheck: @escaping () async -> Swift.Result<Bool, Error>, callback: @escaping (Result) async -> Void) {
```
diegoreymendez, line 55: Just allowing the callback to be async to remove some unnecessary `Task`s.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -263,10 +263,20 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             providerEvents.fire(.rekeyAttempt(.success))
         } catch {
             providerEvents.fire(.rekeyAttempt(.failure(error)))
+            await subscriptionAccessErrorHandler(error)
```
diegoreymendez, line 266: User access errors are handled at the topmost level, instead of within `updateTunnel`.  This is because not all calls to `updateTunnel` need the tunnel to shut down itself.

In fact when starting up the tunnel we want to fail startup returning the relevant error instead.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -587,7 +597,18 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-                self?.handleSettingsChange(change)
+                guard let self else { return }
+                let handleSettingsChange = handleSettingsChange
+                let subscriptionAccessErrorHandler = subscriptionAccessErrorHandler
+
+                Task { @MainActor in
+                    do {
+                        try await handleSettingsChange(change)
+                    } catch {
+                        await subscriptionAccessErrorHandler(error)
```
diegoreymendez, line 608: Just another case where we want to handle subscription access errors.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1236,26 +1245,23 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+        let managers = try await NETunnelProviderManager.loadAllFromPreferences()
 
-            manager.isOnDemandEnabled = false
-            try await manager.saveToPreferences()
-            try await manager.loadFromPreferences()
+        guard let manager = managers.first else {
+            os_log("Could not find a viable manager, bailing out of shutdown", log: .networkProtection)
+            // Doesn't matter a lot what error we throw here, since we'll try cancelling the
+            // tunnel.
+            throw TunnelError.vpnAccessRevoked
```
diegoreymendez, line 1257: If we can't shut down the tunnel, we return an error so the parent method can try to cancel the tunnel.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1236,26 +1245,23 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-            let managers = try await NETunnelProviderManager.loadAllFromPreferences()
-
-            guard let manager = managers.first else {
-                completionHandler?(nil)
-                return
-            }
-
-            os_log("⚪️ Disabling Connect On Demand and shutting down the tunnel", log: .networkProtection)
+    @MainActor
+    public func handleShutDown() async throws {
```
diegoreymendez, line 1249: Making this async so we can remove the task.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1236,26 +1245,23 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            // Doesn't matter a lot what error we throw here, since we'll try cancelling the
+            // tunnel.
+            throw TunnelError.vpnAccessRevoked
+        }
 
-            let error = NSError(domain: "com.duckduckgo.vpn", code: 0)
-            await cancelTunnel(with: error)
+        manager.isOnDemandEnabled = false
+        try await manager.loadFromPreferences()
+        try await manager.saveToPreferences()
```
diegoreymendez, line 1262: I inverted the order of these... no sense in saving and then loading... loading is necessary **before** saving.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1442,9 +1448,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             /// Ignore otherwise
             switch result {
             case .invalidEntitlement:
-                Task { [weak self] in
-                    await self?.handleInvalidEntitlement(attemptsShutdown: true)
-                }
+                await self?.handleAccessRevoked(attemptsShutdown: true)
```
diegoreymendez, line 1451: Renamed and removed task.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1497,14 +1501,23 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-    // Attempt to shut down the tunnel
-    // On iOS 16 and below, as a workaround, we rekey to force a 403 error so that the tunnel fails to restart
     @MainActor
     private func attemptShutdown() async {
+        let cancelTunnel = {
+#if os(macOS)
+            try? self.tokenStore.deleteToken()
+#endif
+            self.cancelTunnelWithError(TunnelError.vpnAccessRevoked)
+        }
```
diegoreymendez, line 1511: Instead of rekeying we do this.  When starting up again the tunnel will rekey on its own if it still has an auth token.

Sources/Common/Concurrency/TaskExtension.swift
```diff
@@ -42,6 +42,29 @@ public extension Task where Success == Never, Failure == Error {
...
+                         cancellationHandler: (@Sendable () async -> Void)? = nil) -> Task {
+
+        Task {
+            do {
+                if let delay {
+                    try await Task<Never, Never>.sleep(interval: delay)
```
samsymons, line 54: Should be able to simplify this down to `Task.sleep` I believe.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1497,14 +1501,23 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
     }
 
-    // Attempt to shut down the tunnel
-    // On iOS 16 and below, as a workaround, we rekey to force a 403 error so that the tunnel fails to restart
     @MainActor
     private func attemptShutdown() async {
+        let cancelTunnel = {
+#if os(macOS)
+            try? self.tokenStore.deleteToken()
+#endif
```
samsymons, line 1509: Do we need to check for system extension vs app extension here? I think it should be fine as-is, but checking anyway.

samsymons (approved):

quanganhdo (approved):

diegoreymendez:
I'll merge this tomorrow, to allow the internal testing week.

Sources/Common/Concurrency/TaskExtension.swift
diegoreymendez: I'll keep it because of this, but not sure why that's even needed. 🤷 

Screenshot 2024-07-30 at 11 57 48 AM

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: Well, if access is revoked it should be fine regardless.  The only risk I see is if we get false positives as access-revoked, but then the problem would still not be this code, if that makes sense.

PR #909, opened 2024-07-24
Improve CrashLogMessageExtractor memory safety

mallexxx:

**Required**:

Task/Issue URL: https://app.asana.com/0/414709148257752/1207878606354585/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3133
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3011

**Description**:
- Move memory protection modifier `withTemporaryUnprotectedMemory` earlier, matching https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
- Add symbol tables/string table regions boundary checks for safer memory access
- Add `swapCxaThrow` parameter to `CrashLogMessageExtractor.setUp` to allow disabling `__cxa_throw` swapping

**Steps to test this PR**:
1. pass `swapCxaThrow: false` to CrashLogMessageExtractor.setUp
2. detach debugger
3. simulate c++ crash
4. run from debugger
5. validate uploaded reports is generated (without the exception stack trace)
6. Simulate NSException and validate the stack trace is there

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/Crashes/CxaThrowSwapper.swift
+++ b/Sources/Crashes/CxaThrowSwapper.swift
@@ -112,7 +112,7 @@ private func processMachHeader(_ header: UnsafePointer<mach_header>?, slide: Int
 }
 
 private var cxxOriginalThrowFunctions = [UnsafeRawPointer: UnsafeRawPointer]()
-private let indicesToSkip = [UInt32(bitPattern: INDIRECT_SYMBOL_ABS), INDIRECT_SYMBOL_LOCAL, UInt32(Int64(INDIRECT_SYMBOL_LOCAL) | Int64(INDIRECT_SYMBOL_ABS))]
+private let indicesToSkip = [UInt32(INDIRECT_SYMBOL_ABS), INDIRECT_SYMBOL_LOCAL, INDIRECT_SYMBOL_LOCAL | UInt32(INDIRECT_SYMBOL_ABS)]
 
 private func _processMachHeader(_ header: UnsafePointer<mach_header_64>?, slide: Int) throws {
     guard let header, slide != 0 else { throw ProcessingError.zeroSlide }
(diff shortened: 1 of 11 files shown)


Sources/Crashes/CxaThrowSwapper.swift
```diff
@@ -121,23 +121,32 @@ private func _processMachHeader(_ header: UnsafePointer<mach_header_64>?, slide:
...
+        where [S_LAZY_SYMBOL_POINTERS, S_NON_LAZY_SYMBOL_POINTERS].contains(section.pointee.type) {
+
+            try section.pointee.indirectSymbolBindings(slide: slide)?.withTemporaryUnprotectedMemory { indirectSymbolBindings in
+                guard let indirectSymbolIndices = section.pointee.indirectSymbolIndices(indirectSymtab: imageMap.indirectSymtab) else { return }
+                for i in 0..<section.pointee.count where indirectSymbolIndices.indices.contains(i) && indirectSymbolBindings.indices.contains(i) {
+                    let symtabIndex = indirectSymbolIndices[i]
+                    guard !indicesToSkip.contains(symtabIndex),
+                          let symbolName = imageMap.symbolName(at: Int(symtabIndex)),
+                          symbolName[0] != 0, symbolName[1] != 0,
+                          strcmp(symbolName.advanced(by: 1), "__cxa_throw") == 0 else { continue }
```
bwaresiak, line 140: TBH I'm quite curious why we advance by 1 here :)

Sources/Crashes/CxaThrowSwapper.swift
```diff
@@ -112,7 +112,7 @@ private func processMachHeader(_ header: UnsafePointer<mach_header>?, slide: Int
 }
 
 private var cxxOriginalThrowFunctions = [UnsafeRawPointer: UnsafeRawPointer]()
-private let indicesToSkip = [UInt32(bitPattern: INDIRECT_SYMBOL_ABS), INDIRECT_SYMBOL_LOCAL, UInt32(Int64(INDIRECT_SYMBOL_LOCAL) | Int64(INDIRECT_SYMBOL_ABS))]
+private let indicesToSkip = [UInt32(INDIRECT_SYMBOL_ABS), INDIRECT_SYMBOL_LOCAL, INDIRECT_SYMBOL_LOCAL | UInt32(INDIRECT_SYMBOL_ABS)]
```
bwaresiak, line 115: I winder if we can have some tests that find these, feed them int the logic and validate we correctly skip these.

bwaresiak (commented):
@mallexxx I went through these changes, these looks safer programmatically but it's getting more and more difficult to process what's going on here. :) 

So two questions:
1. How we can make this more accessible to regular iOS/macOS dev that typically sees dyld only during WWDC highlights :) 
2. Can we test this handler logic?

Sources/Crashes/Mach-O/ImageMap.swift
```diff
@@ -77,12 +80,13 @@ struct ImageMap {
...
-        let strtabOffset = symtab[symtabIndex].n_un.n_strx
-        let symbolName = strtab.advanced(by: Int(strtabOffset))
-        let symbolNameStr = String(cString: symbolName)
-
-        return symbolNameStr
+    func symbolName(at symtabIndex: Int) -> UnsafePointer<CChar>? {
+        guard symtab.indices.contains(symtabIndex) else { return nil }
+        let strtabOffset = Int(symtab[symtabIndex].n_un.n_strx)
+        guard strtab.indices.contains(strtabOffset),
+              strtab.indices.contains(strtabOffset + 1) else { return nil }
```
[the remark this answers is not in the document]
mallexxx: there were crashes in swift‘s `String(cString:)` initializer for some C string pointers, I‘ve changed it to match the original `KSCrash` code using `strcmp` (bytewise compare) that doesn‘t copy the memory and stop at the first non-matching byte, so the check is there from `KSCrash`, probably `strcmp` would just work…

Sources/Crashes/CxaThrowSwapper.swift
```diff
@@ -121,23 +121,32 @@ private func _processMachHeader(_ header: UnsafePointer<mach_header_64>?, slide:
...
+                    let symtabIndex = indirectSymbolIndices[i]
+                    guard !indicesToSkip.contains(symtabIndex),
+                          let symbolName = imageMap.symbolName(at: Int(symtabIndex)),
+                          symbolName[0] != 0, symbolName[1] != 0,
+                          strcmp(symbolName.advanced(by: 1), "__cxa_throw") == 0 else { continue }
+
+                    let sectionInfo = try Dl_info(section)
+                    os_log(.debug, log: CxaThrowSwapper.log, "found %s: %s", String(cString: symbolName), indirectSymbolBindings[i].debugDescription)
 
                     cxxOriginalThrowFunctions[UnsafeRawPointer(sectionInfo.dli_fbase)] = indirectSymbolBindings[i]
```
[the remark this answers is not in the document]
mallexxx: in general this should catch all the throws, write the last exception stack trace into a Thread dictionary and pass control to the designated handler (not inevitably leading to a crash), worths adding an extra C++ test but should just work.

Sources/Crashes/CxaThrowSwapper.swift
mallexxx: me too :)
it‘s skipping the first `_`, that‘s the original approach in `KSCrash`

Sources/Crashes/CxaThrowSwapper.swift
mallexxx: yes, let‘s plan a follow-up task to add some tests for the `__cxa_throw` hook machinery and for now release it disabled both for iOS and macOS

Sources/Crashes/CxaThrowSwapper.swift
ayoy: If you check symbol table for a binary (e.g. `nm /Applications/DuckDuckGo.app/Contents/MacOS/DuckDuckGo | less`) then you'll notice that all symbols have a `_` prefix. Swift mangled function names are prefixed with an underscore as well, and `__cxa_throw` is also prefixed with `_` :)

ayoy (approved):
This looks good to me, great job @mallexxx! I appreciate the recent changes to sanitizing the crash log 👏 

For more context to others reading it (cc @bwaresiak):
* I've spent some time playing with this last night, going through the code and trying to make sense of it and it worked quite well with the help of ChatGPT. I've synced with Alex on MM this morning and we talked about adding documentation (even though the code is a rewrite of KSCrash, it's now our code, so better be clear about what happens there and why). The docs are added now so it's good to go 💪 
* I agree it's a good idea to temporarily disable __cxa_throw swapping and re-enable it after adding tests. I've also spent some time last night trying to come up with a test, with a moderate success, then talked to Alex and he confirmed my rough plan, but I'll anyway leave it with Alex to implement tests.

PR #924, opened 2024-08-05
VPN snooze mode

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1207974416599035/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3184
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3085

**Description**:

This PR adds VPN snooze mode support.

**Steps to test this PR**:
1. See client PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1670,6 +1690,97 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                return
+            }
+
+            if error == nil {
+                self.connectionStatus = .snoozing
+                self.snoozeTimingStore.activeTiming = .init(startDate: Date(), duration: duration)
+                self.notificationsPresenter.showSnoozingNotification(duration: duration)
+
+                let timer = DispatchSource.makeTimerSource(queue: self.timerQueue)
+                timer.schedule(deadline: .now() + duration, repeating: .never)
```
samsymons, line 1738: I need to change the timer implementation to use a Task rather than a DispatchSource because the DispatchSource is known to have issues inside a VPN extension.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1670,6 +1690,97 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+                        await self.cancelSnooze()
+                    }
+
+                    timer.cancel()
+                }
+
+                self.snoozeTimer = timer
+                self.snoozeTimer?.resume()
+            } else {
+                self.snoozeTimingStore.reset()
```
quanganhdo, line 1751: Shouldn't we do something here, like firing an error pixel?

quanganhdo (approved):

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: The one error that can occur is if the adapter is already stopped, which is an error we're not too concerned about; it could happen if a snooze call comes in right as the VPN is shutting down, which isn't too interesting. I'll likely come back and add a pixel here to be safe though, after triaging it first.

PR #932, opened 2024-08-08
Add Phishing Detection Data Manager

not-a-rootkit:

**Required**:

Task/Issue URL: https://app.asana.com/0/1204023833050360/1207976613228504/f
iOS PR: -
macOS PR: -

**Description**:
Introduces data related classes for the phishing detection feature, specifically:
1. Embedded data set provider
2. Background activity scheduler
3. Data types and data store 

Note: dataUpdater and API client is implemented in a separate PR. 

**Steps to test this PR**:
1. Run BackgroundActivityScheduler tests 

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/PhishingDetection/PhishingDetectionDataStore.swift
```diff
@@ -0,0 +1,134 @@
...
+                throw PhishingDetectionDataError.empty
+            }
+        } catch {
+            os_log(.debug, log: .phishingDetection, "\(self): 🔴 Error loading phishing protection data: \(error). Reloading from embedded dataset.")
+            self.currentRevision = dataProvider.embeddedRevision
+            self.hashPrefixes = dataProvider.loadEmbeddedHashPrefixes()
+            self.filterSet = dataProvider.loadEmbeddedFilterSet()
+        }
+    }
+}
```
SabrinaTardio, line 134: I think it would be nice to separate some responsabilities and have a FileStorageManager
`lass FileStorageManager {
    let dataStoreURL: URL

    init(dataStoreURL: URL) {
        self.dataStoreURL = dataStoreURL
        createDirectoryIfNeeded()
    }

    private func createDirectoryIfNeeded() {
        do {
            try FileManager.default.createDirectory(at: dataStoreURL, withIntermediateDirectories: true, attributes: nil)
        } catch {
            os_log(.error, log: .default, "Failed to create directory: %{public}@", error.localizedDescription)
        }
    }

    func write(data: Data, to filename: String) {
        let fileURL = dataStoreURL.appendingPathComponent(filename)
        do {
            try data.write(to: fileURL)
        } catch {
            os_log(.error, log: .default, "Error writing to %{public}@: %{public}@", filename, error.localizedDescription)
        }
    }

    func read(from filename: String) -> Data? {
        let fileURL = dataStoreURL.appendingPathComponent(filename)
        do {
            return try Data(contentsOf: fileURL)
        } catch {
            os_log(.error, log: .default, "Error reading from %{public}@: %{public}@", filename, error.localizedDescription)
            return nil
        }
    }
}`

something like this

Sources/PhishingDetection/PhishingDetectionDataStore.swift
```diff
@@ -0,0 +1,134 @@
...
+
+            try hashPrefixesData.write(to: hashPrefixesFileURL)
+            try filterSetData.write(to: filterSetFileURL)
+            try revision.write(to: revisionFileURL)
+        } catch {
+            os_log(.debug, log: .phishingDetection, "\(self): 🔴 Error saving phishing protection data: \(error)")
+        }
+    }
+
+    public func loadData() async {
```
SabrinaTardio, line 109: Just some ideas:

maybe on loadData we could just write the embedded data to the file URL if none available and we could make the variables lazy variable
Something like this:

   lazy var filterSet: Set<Filter> = {
        loadSet(from: "filterSet.json") ?? dataProvider.loadEmbeddedFilterSet()
    }() 
    
    (and then you update them from the write func

Sources/PhishingDetection/PhishingDetectionDataStore.swift
```diff
@@ -0,0 +1,134 @@
...
+
+            hashPrefixes = Set(try decoder.decode([String].self, from: hashPrefixesData))
+            filterSet = Set(try decoder.decode([Filter].self, from: filterSetData))
+            currentRevision = try decoder.decode(Int.self, from: revisionData)
+
+            if (hashPrefixes.isEmpty && filterSet.isEmpty) || currentRevision == 0 {
+                throw PhishingDetectionDataError.empty
+            }
+        } catch {
+            os_log(.debug, log: .phishingDetection, "\(self): 🔴 Error loading phishing protection data: \(error). Reloading from embedded dataset.")
```
SabrinaTardio, line 128: I am not sure cause the very first time it should not be an error and it’s expected. It would also be an error if we do something like I suggested above

SabrinaTardio (commented):
I left a few comments. also I assume the tests for the other classes will be in follow up PRs.
Some changes are suggestions idea but we should aim to merge this part in the main branch today and move on from that


Let me know if you have any questions!

Sources/PhishingDetection/PhishingDetectionDataActivities.swift
```diff
@@ -0,0 +1,115 @@
...
+    init(interval: TimeInterval, identifier: String) {
+        self.interval = interval
+        self.identifier = identifier
+    }
+
+    func start(activity: @escaping () async -> Void) {
+        stop()
+        task = Task {
+            let taskId = UUID().uuidString
+            while true {
```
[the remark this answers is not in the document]
not-a-rootkit: I think you're right - I wanted to do this originally actually, not sure how I ended up with the while-true.

Sources/PhishingDetection/PhishingDetectionDataProvider.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    public func loadEmbeddedFilterSet() -> Set<Filter> {
+        do {
+            let filterSetData = try Data(contentsOf: embeddedFilterSetURL)
+            let sha256 = SHA256.hash(data: filterSetData)
+            let hashString = sha256.compactMap { String(format: "%02x", $0) }.joined()
+
+            guard hashString == embeddedFilterSetDataSHA else {
+                os_log(.debug, log: .phishingDetection, "\(self): 🔴 Fatal Error: SHA mismatch for filterSet JSON file. Expected \(embeddedFilterSetDataSHA), got \(hashString)")
+//                assertionFailure("SHA mismatch for filterSet JSON file. Expected \(embeddedFilterSetDataSHA), got \(hashString)")
```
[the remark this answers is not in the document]
not-a-rootkit: Probably not - I think it is safer to just log an error and return empty set so the feature is effectively disabled, rather than crash the whole app.

SabrinaTardio (approved):
Great job!
Just  there are two warnings… feel free to merge when sorted

PR #947, opened 2024-08-20
Make remote config accessible to background agents

SlayterDev:

**Required**:

Task/Issue URL: https://app.asana.com/0/1203581873609357/1207165680693234/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3255
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3124

**Optional**:

Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1.
1.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/Configuration/DefaultConfigurationManager.swift
+++ b/Sources/Configuration/DefaultConfigurationManager.swift
@@ -77,8 +77,7 @@ open class DefaultConfigurationManager {
     public var lastRefreshCheckTime: Date = Date()
 
     public func start() {
-        // TODO: Move .config log to BSK
-        os_log("Starting configuration refresh timer", log: .default, type: .debug)
+        os_log("Starting configuration refresh timer", log: .config, type: .debug)
         timerCancellable = Timer.publish(every: Constants.refreshCheckIntervalSeconds, on: .main, in: .default)
             .autoconnect()
             .receive(on: Self.queue)
@@ -102,7 +101,7 @@ open class DefaultConfigurationManager {
     @discardableResult
     private func refreshIfNeeded() -> Task<Void, Never>? {
         guard isReadyToRefresh else {
-            os_log("Configuration refresh is not needed at this time", log: .default, type: .debug)
+            os_log("Configuration refresh is not needed at this time", log: .config, type: .debug)
             return nil
         }
         return Task {
--- a/Sources/Configuration/DefaultConfigurationManager.swift
+++ b/Sources/Configuration/DefaultConfigurationManager.swift
@@ -50,23 +50,28 @@ open class DefaultConfigurationManager {
         public static let retryDelaySeconds = 60.0 * 60 * 1 // 1 hour delay before checking again if something went wrong last time
         public static let refreshCheckIntervalSeconds = 60.0 // check if we need a refresh every minute
 
+        static let lastUpdateDefaultsKey = "configuration.lastUpdateTime"
     }
+
+    private var defaults: UserDefaults
 
     public var fetcher: ConfigurationFetcher
 
-    public init(fetcher: ConfigurationFetcher) {
+    public init(fetcher: ConfigurationFetcher, defaults: UserDefaults = UserDefaults()) {
         self.fetcher = fetcher
-
-        self.lastUpdateTime = .distantPast
+        self.defaults = defaults
     }
 
     public static let queue: DispatchQueue = DispatchQueue(label: "Configuration Manager")
 
-//    @UserDefaultsWrapper(key: .configLastUpdated, defaultValue: .distantPast)
-    private(set) var lastUpdateTime: Date
-
-//    @UserDefaultsWrapper(key: .configLastInstalled, defaultValue: nil)
-    private(set) var lastConfigurationInstallDate: Date?
+    open internal(set) var lastUpdateTime: Date {
+        get {
+            defaults.object(forKey: Constants.lastUpdateDefaultsKey) as? Date ?? .distantPast
+        }
+        set {
+            defaults.set(newValue, forKey: Constants.lastUpdateDefaultsKey)
+        }
+    }
 
     private var timerCancellable: AnyCancellable?
     public var lastRefreshCheckTime: Date = Date()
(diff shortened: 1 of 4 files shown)


Tests/BrowserServicesKitTests/Resources/privacy-reference-tests
samsymons, line 1: We can revert this change back to whatever is on `main`.

Sources/Configuration/DefaultConfigurationManager.swift
```diff
@@ -0,0 +1,160 @@
...
+        Task {
+            await refreshNow(isDebug: isDebug)
+        }
+    }
+
+    /// Will try to update the config again at the regularly scheduled interval
+    /// **Note:** You must call `start()` on your `ConfigurationManager` instance for this to take effect. It relies on the internal refresh loop of the
+    /// `DefaultConfigurationManager` class
+    public func tryAgainLater() {
+        lastUpdateTime = Date()
```
samsymons, line 135: I'll need to check how we use this, right now it's not thread safe but I understand that it should always be called from a specific queue - maybe we can use a dispatch assertion. I'll take a look at the clients and report back.

Tests/ConfigurationTests/ConfigurationManagerTests.swift
```diff
@@ -0,0 +1,201 @@
...
+@testable import Networking
+@testable import TestUtils
+
+final class MockConfigurationManager: DefaultConfigurationManager {
+
+    var dependencyProvider: MockDependencyProvider = MockDependencyProvider()
+    var name: String?
+
+    override func refreshNow(isDebug: Bool = false) async {
+        let configFetcched = await fetchConfigDependencies(isDebug: isDebug)
```
samsymons, line 31:
```suggestion
        let configFetched = await fetchConfigDependencies(isDebug: isDebug)
```

Tests/ConfigurationTests/ConfigurationManagerTests.swift
```diff
@@ -0,0 +1,201 @@
...
+        updateConfigDependencies()
+    }
+}
+
+struct MockDependencyProvider {
+    var privacyConfigEtag: String?
+    var privacyConfigData: Data?
+}
+
+final class MockDefaults: KeyValueStoring {
```
samsymons, line 71: We can reuse `MockKeyValueStore` from the `TestUtils` package here 😄

Tests/ConfigurationTests/ConfigurationManagerTests.swift
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    func testWhenManagerAIsUpdatedManagerBIsAlsoUpdated() async throws {
+        let managerA = makeConfigurationManager(name: "A")
+        let managerB = makeConfigurationManager(name: "B")
+
+        let configData = Data("Privacy Config".utf8)
+        MockURLProtocol.requestHandler = { _ in (HTTPURLResponse.ok, configData) }
+        await managerA.refreshNow()
+        try await Task.sleep(interval: 2)
```
samsymons, line 149: There may be a way we can instead observe the changes we're waiting for, and shortcut the need to sleep - I'll think on it a bit.

Sources/Configuration/DefaultConfigurationManager.swift
```diff
@@ -0,0 +1,160 @@
...
+        timerCancellable = Timer.publish(every: Constants.refreshCheckIntervalSeconds, on: .main, in: .default)
+            .autoconnect()
+            .receive(on: Self.queue)
+            .sink(receiveValue: { _ in
+                self.lastRefreshCheckTime = Date()
+                self.refreshIfNeeded()
+            })
+        Task {
+            await refreshNow()
+        }
```
samsymons, line 103: `NSTimer` won't keep time consistently in a network extension environment (Apple forum thread about it [here](https://forums.developer.apple.com/forums/thread/687170)), but given that we don't need perfect timing I'd say it's acceptable. What do you think @diegoreymendez?

Tests/ConfigurationTests/ConfigurationManagerTests.swift
```diff
@@ -0,0 +1,207 @@
...
+                                               defaults: sharedDefaults)
+        manager.name = name
+        return manager
+    }
+
+    func testWhenConfigIsFetchedAndStoredDependencyIsUpdated() async {
+        let configurationManager = makeConfigurationManager()
+
+        let configData = Data("Privacy Config".utf8)
+        MockURLProtocol.requestHandler = { _ in (HTTPURLResponse.ok, configData) }
```
samsymons, line 107: `MockURLProtocol` is static but isn't resetting itself between tests so I don't trust that we won't get inconsistent behaviour when running the entire test suite - would you be able to add a `tearDown` function to it and call it after running tests? Something like this:

```
static func tearDown() {
    lastRequest = nil
    requestHandler = nil
}
```

Tests/ConfigurationTests/ConfigurationManagerTests.swift
```diff
@@ -0,0 +1,207 @@
...
+        await managerA.refreshNow()
+        await fulfillment(of: [e!], timeout: 2)
+
+        XCTAssertEqual(managerB.dependencyProvider.privacyConfigData, configData)
+        XCTAssertEqual(managerB.dependencyProvider.privacyConfigEtag, HTTPURLResponse.testEtag)
+
+        MockURLProtocol.requestHandler = { _ in (HTTPURLResponse.internalServerError, nil) }
+        await managerB.refreshNow()
```
samsymons, line 201: Just to be safe we could also assert against the `lastRequest` property of `MockURLProtocol`, to prove for sure that the request is being made and thus the behavior is correct.

samsymons (approved):
Approving with some comments, it works well for me but there are a few areas for improvement, like adding a test suite to the iOS manager and supporting embedded config files on the non-app targets.

diegoreymendez (approved):

PR #964, opened 2024-08-27
Alesab/onboarding

SabrinaTardio:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1204186595873227/1208077416568671/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3282
macOS PR:  https://github.com/duckduckgo/macos-browser/pull/3149

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1208094950223874/f

**Description**:  Extracts code from the onboarding that can be used by macOS as well

**Steps to test this PR**:
1. Test new onboarding in iOS
1. Look at ContextualDaxDialogContent previews

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Package.swift
```diff
@@ -397,6 +397,9 @@ let package = Package(
             dependencies: [
                 "BrowserServicesKit"
             ],
+            resources: [
+                .process("Resources")
```
alessandroboron, line 401: That’s interesting, according to Apple this shouldn’t be needed 🤔 

```
When you add a resource to your Swift package, Xcode detects common resource types for Apple platforms and treats them as a resource automatically. For example, you don’t need to make changes to your package manifest for the following resources:
Interface Builder files; for example, XIB files and storyboards
Core Data files; for example, xcdatamodeld files
Asset catalogs
.lproj folders you use to provide localized resources
If you add a resource file that Xcode doesn’t treat as a resource by default, you must configure it in your package manifest, as described in the next section.
```

alessandroboron (approved):
LGTM! I couldn’t check the previews on BSK because the PreviewXCAgent crashed for me. But I tested iOS and it looks good to me!

I think BSK needs a `minor` bump instead of `patch` as we added a new library.

Package.swift
SabrinaTardio: Yeah… apparently macOS does not like it without 🥲

PR #966, opened 2024-08-29
Fix site owned by major network

alessandroboron:

**Required**:

Task/Issue URL: https://app.asana.com/0/1176956903599313/1208164650000833/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3286
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3168
What kind of version bump will this require? Minor

**Description**:

There’s an existing [issue](https://app.asana.com/0/0/1207970856546695/1208097565947206/f) that causes showing the wrong education dialog for websites that are owned by a major network like Facebook and Google.

In addition, the Privacy Dashboard [does not reflect weather an entity is owned by a major tracker](https://app.asana.com/0/0/1207970856546695/1208114308033513/f)

This PR exposes the field “ownedBy” with the purpose of identifying when a tracker is owned by another entity. e.g. Instagram is ownedBy is Facebook

**Steps to test this PR**:
1. See iOS  
1. See macOS

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/ContentBlocking/TrackerDataQueryExtension.swift
+++ b/Sources/BrowserServicesKit/ContentBlocking/TrackerDataQueryExtension.swift
@@ -34,10 +34,10 @@ extension TrackerData {
         return nil
     }
 
-    /// Returns the entity associated with the host. If the entity is owned by a parent entity, it returns the parent entity.
+    /// Returns the parent of the entity associated with the host if any. If the entity associated with the host doesn't have a parent return the entity. Otherwise, return nil.
     /// - Parameter host:
     /// - Returns: The entity associated with the host.
-    public func findParentOrFallback(forHost host: String) -> Entity? {
+    public func findParentEntityOrFallback(forHost host: String) -> Entity? {
         // If the entity associated with the host is owned by a parent company (e.g. Instagram is owned by Facebook) return the parent company.
         // If the entity associated with the host is not owned by a parent company return the entity.
         // If the are no entities associated with the host return nil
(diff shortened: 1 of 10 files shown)


Sources/BrowserServicesKit/ContentBlocking/TrackerDataQueryExtension.swift
```diff
@@ -34,6 +34,23 @@ extension TrackerData {
         return nil
     }
 
+    /// Returns the parent of the entity associated with the host if any. If the entity associated with the host doesn't have a parent return the entity. Otherwise, return nil.
+    /// - Parameter host:
+    /// - Returns: The entity associated with the host.
+    public func findParentEntityOrFallback(forHost host: String) -> Entity? {
```
alessandroboron, line 40: Perhaps I can just modify instead of creating a new one. 

```
public func findEntity(forHost host: String) -> Entity? {
        for host in variations(of: host) {
            if let entityName = domains[host] {
                return entities[entityName]
            }
        }
        return nil
    }

```

`TrackerResolver` & `ReferrerTrimming` use `findEntity(forHost` and I’m not sure if we always want to find the parent entity if it exists. I think this is safer?

brindy (approved):
LGTM! Nice work

alessandroboron:
> LGTM! Nice work

Thanks! Just FYI, initially I thought that this required a minor bump. Because the TrackerRadarKit change adds a variable in the init for Owner, this will be a major change as it break client tests that instantiate Owner(s).

PR #987, opened 2024-09-11
Networking v2

federicocappelli:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207991044706235/f
iOS PR: Not adopted yet
macOS PR: Not adopted yet

Tech Design URL: https://app.asana.com/0/414235014887631/1208247010475423/f
CC:

**Description**:

During the investigations for the main task, I realised that an update to the networking library was necessary to allow correct testability and adherence to the best coding standards. this is Networking V2 that will be used, first in the above task and then for any other feature that needs to execute an API request.

Please refer to the Networking [README](https://github.com/duckduckgo/BrowserServicesKit/blob/a11b58b004358a85e6597a8f30c9cc85119f83dc/Sources/Networking/README.md) for details and to the Unit tests for examples

**Steps to test this PR**:
- Run unit tests in `BSK/Tests/NetworkingTests/v2`
- Note: some tests are disabled because they execute real network requests, they can be useful for experiments and learning.

---
###### Internal references:


Sources/Networking/v1/APIRequestConfiguration.swift
```diff
@@ -24,23 +24,21 @@ extension APIRequest {
...
 
         public init(url: URL,
-                    method: HTTPMethod = .get,
+                    method: APIRequest.HTTPMethod = .get,
                     queryParameters: QueryParams = [],
                     allowedQueryReservedCharacters: CharacterSet? = nil,
                     headers: APIRequest.Headers = APIRequest.Headers(),
                     body: Data? = nil,
                     timeoutInterval: TimeInterval = 60.0,
-                    attribution: URLRequestAttribution? = .developer,
```
federicocappelli, line 43: URLRequestAttribution was never used in any project and the apple default is .developer already, re-setting is redundant.

Sources/Networking/README.md
```diff
@@ -0,0 +1,66 @@
+#  Networking
+
+This is the preferred Networking library for iOS and macOS DuckDuckGo apps.
+If the library doesn't have the features you require, please improve it. 
+
+## v2
+
+### USage
```
aataraxiaa, line 8:
```suggestion
### Usage
```

Sources/Networking/v2/APIRequestV2.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public var debugDescription: String {
+        """
+        \(urlRequest.httpMethod ?? "Nil") \(urlRequest.url?.absoluteString ?? "nil")
+        Headers: \(urlRequest.allHTTPHeaderFields?.debugDescription ?? "-")
+        Body: \(urlRequest.httpBody?.debugDescription ?? "-")
+        Constraints: \(responseConstraints?.debugDescription ?? "-")
+        """
+    }
```
aataraxiaa, line 77: Nice idea to add this. While we are here, maybe we can add all properties, i.e:

```suggestion
   public var debugDescription: String {
        """
        APIRequestV2:
        URL: \(urlRequest.url?.absoluteString ?? "nil")
        Method: \(urlRequest.httpMethod ?? "Nil")
        Headers: \(urlRequest.allHTTPHeaderFields?.debugDescription ?? "-")
        Body: \(urlRequest.httpBody?.debugDescription ?? "-")
        Timeout Interval: \(timeoutInterval)s
        Cache Policy: \(urlRequest.cachePolicy)
        Response Constraints: \(responseConstraints?.map { $0.rawValue } ?? [])
        """
    }
```

Sources/Networking/v2/APIRequestV2.swift
```diff
@@ -0,0 +1,78 @@
...
+        guard let finalURL = urlComps.url else {
+            return nil
+        }
+        var request = URLRequest(url: finalURL, timeoutInterval: timeoutInterval)
+        request.allHTTPHeaderFields = headers?.httpHeaders
+        request.httpMethod = method.rawValue
+        request.httpBody = body
+        if let cachePolicy = cachePolicy {
+            request.cachePolicy = cachePolicy
+        }
```
aataraxiaa, line 66: To keep the init tidy (and easily readable), maybe we could extract request creation to a `buildURLRequest` method?

Sources/Networking/v2/APIResponseConstraints.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public enum APIResponseConstraints: String, CustomDebugStringConvertible {
```
aataraxiaa, line 21: 👍🏼I think this is a more descriptive name than the existing `APIResponseRequirements`

Sources/Networking/v2/APIService.swift
```diff
@@ -0,0 +1,126 @@
...
+    typealias APIResponse = (data: Data?, httpResponse: HTTPURLResponse)
+
+    func fetch<T: Decodable>(request: APIRequestV2) async throws -> T?
+    func fetch(request: APIRequestV2) async throws -> APIService.APIResponse
+}
+
+public struct DefaultAPIService: APIService {
+    private let urlSession: URLSession
+
+    public init(urlSession: URLSession = .shared) {
```
aataraxiaa, line 33: We should consider also injecting the `JSONDecoder`. This might be useful if we want to change decoding attributes. e.g for APIs with snake case:

```
let decoder = JSONDecoder()
decoder.keyDecodingStrategy = .convertFromSnakeCase
```

Sources/Networking/v2/APIService.swift
```diff
@@ -0,0 +1,126 @@
...
+            guard let resultString = String(data: data, encoding: .utf8) as? T else {
+                let error = APIRequestV2.Error.invalidDataType
+                Logger.networking.error("Error: \(error.localizedDescription)")
+                throw error
+            }
+            return resultString
+        default:
+            // Decode data
+            let decoder = JSONDecoder()
+            return try decoder.decode(T.self, from: data)
```
aataraxiaa, line 63: Although we are switching on the type (strong -vs- other), we are really just decoding in both cases. So, maybe we could extract all of this to a `decode` method. This method could throw an error which we then handle and throw as `APIRequestV2.Error.invalidDataType`. This would mean we are consistent in throwing `APIRequestV2.Error.invalidDataType` when decoding fails for either string or other types. e.g:

```
  private func decode<T: Decodable>(data: Data, to type: T.Type) throws -> T {
        if T.self == String.self, let resultString = String(data: data, encoding: .utf8) as? T {
            return resultString
        } else {
            return try decoder.decode(T.self, from: data)
        }
    }
```

and then the callsite is 

```
 do {
            return try decode(data: data, to: T.self)
        } catch {
            Logger.networking.error("Error decoding data: \(error.localizedDescription)")
            throw APIRequestV2.Error.invalidDataType
        }
```

Sources/Networking/v2/APIService.swift
```diff
@@ -0,0 +1,126 @@
...
+                    guard let userAgent = httpResponse.allHeaderFields[HTTPHeaderKey.userAgent] as? String,
+                          userAgent.isEmpty == false else {
+                        let error = APIRequestV2.Error.unsatisfiedRequirement(requirement)
+                        Logger.networking.error("Error: \(error.localizedDescription)")
+                        throw error
+                    }
+                default: break
+                }
+            }
+        }
```
aataraxiaa, line 111: Maybe we could extract this to another `validateResponse` method.

aataraxiaa (commented):
Nice work @federicocappelli , left some comments.

Sources/Networking/v2/APIRequestV2.swift
federicocappelli: Sure! thanks

Sources/Networking/v2/APIRequestV2.swift
federicocappelli: Being `self.urlRequest` a let I can't move the variable init in another function, plus the `buildURLRequest` would need almost all parameters that are in the init, I preferred not to add all this boilerplate code and not to change urlRequest in a var optional just for code clarity. The init is supposed to fail if the urlrequest can't be created.

Sources/Networking/v2/APIService.swift
federicocappelli: This and the previous comments are on point, I'm actually thinking of removing this call completely and leaving the decoding to the user. I added this call as a utility but I have never used it, I always need the HTTPURLResponse for checking response codes and headers.

aataraxiaa: Your call. If we don’t include decoding now we won’t be prevented from adding it later.

Sources/Networking/v2/APIService.swift
miasma13: +1 for that suggestion, snake case and date time formats are prime examples where it may be handy.

Sources/Networking/v2/APIRequestV2.swift
```diff
@@ -0,0 +1,82 @@
...
+        self.urlRequest = request
+    }
+
+   public var debugDescription: String {
+        """
+        APIRequestV2:
+        URL: \(urlRequest.url?.absoluteString ?? "nil")
+        Method: \(urlRequest.httpMethod ?? "Nil")
```
miasma13, line 74: Nil -> nil

Sources/Networking/v2/APIService.swift
federicocappelli: I agree, done

Sources/Networking/v2/APIService.swift
federicocappelli: So, I changed the approach completely, turns out that doing slightly more advanced stuff with tuples is a pain, so:

1. I created a new `APIResponseV2` struct and removed the APIResponse tuple, the content is the same
2. I removed `fetch<T: Decodable>(request: APIRequestV2) async throws -> T?`
3. I created a `APIResponseV2` extension with a function `func decodeBody<T: Decodable>(decoder: JSONDecoder = JSONDecoder()) throws -> T` use:

```
let response = try await apiService.fetch(request: request)
let responseHTML: String = try response.decodeBody()
```
Much more clean and simple + the decoder can be injected, as suggested

Sources/Networking/v2/APIRequestV2.swift
federicocappelli: done, thanks

Sources/Networking/v2/APIResponseV2.swift
```diff
@@ -0,0 +1,51 @@
...
+        switch T.self {
+        case is String.Type:
+            guard let resultString = String(data: data, encoding: .utf8) as? T else {
+                let error = APIRequestV2.Error.invalidDataType
+                Logger.networking.error("Error: \(error.localizedDescription)")
+                throw error
+            }
+            return resultString
+        default:
+            return try decoder.decode(T.self, from: data)
```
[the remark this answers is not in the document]
federicocappelli: I think propagating the original decoding error is the right approach here, we don't want to miss the decoding failure reason and `invalidDataType` may not be the failure reason

aataraxiaa (approved):
LGTM @federicocappelli! A good start to Networking v2. 👍🏼

PR #998, opened 2024-09-19
Add tabs to suggestions/autocomplete on iOS

brindy:
**Required**:

Task/Issue URL: https://app.asana.com/0/392891325557410/1208036752520497/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3371
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3309

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1208270496320813/f
CC: 

**Description**:
Adds support for tabs to suggestions.

**Steps to test this PR**:
1. See platform specific PRs


Sources/Suggestions/Score.swift
```diff
@@ -39,7 +40,9 @@ extension Score {
             score += 300
             // Prioritize root URLs most
             if url.isRoot { score += 2000 }
-        } else if lowercasedTitle.starts(with: query) {
+        } else if lowercasedTitle
+                .trimmingCharacters(in: .alphanumerics.inverted)
```
brindy, line 44: This fixes a bug where if a title was like `"Cats and Dogs"` Cats would not get matched at all because of the quote.

Sources/Suggestions/Score.swift
```diff
@@ -26,6 +26,7 @@ extension Score {
     // swiftlint:disable:next cyclomatic_complexity
     init(title: String?, url: URL, visitCount: Int, query: Query, queryTokens: [Query]? = nil) {
         // To optimize, query tokens can be precomputed
+        let query = query.lowercased()
```
brindy, line 29: Probably not strictly required because usually input will be lowercase, but while testing caused some confusion.

Sources/Suggestions/Score.swift
```diff
@@ -104,12 +111,9 @@ extension Score {
     }
 
     static func tokens(from query: Query) -> [Query] {
-        return query
-            .split(whereSeparator: {
-                $0.unicodeScalars.contains(where: { CharacterSet.whitespacesAndNewlines.contains($0) })
```
brindy, line 109: This seemed superfluous and appeared to result in some strings not tokenising as expected, so this is aligned with the similar code for Bookmarks on iOS.

Sources/Suggestions/SuggestionProcessing.swift
```diff
@@ -49,21 +58,23 @@ final class SuggestionProcessing {
...
         // Combine HaB and domains into navigational suggestions and remove duplicates
         let navigationalSuggestions = allLocalSuggestions + duckDuckGoDomainSuggestions
 
         let maximumOfNavigationalSuggestions = min(
             Self.maximumNumberOfSuggestions - Self.minimumNumberInSuggestionGroup,
             query.count + 1)
-        let mergedSuggestions = merge(navigationalSuggestions, maximum: maximumOfNavigationalSuggestions)
+        let expandedSuggestions = replaceHistoryWithBookmarksAndTabs(navigationalSuggestions)
+
+        let dedupedNavigationalSuggestions = Array(dedupSuggestions(expandedSuggestions).prefix(maximumOfNavigationalSuggestions))
```
brindy, line 71: Should behave the same as before but split out to make reasoning a bit easier.  Effectively history items are replaced by the first matching bookmark and/or tab that is found, but otherwise the list is not changed.  Then we run a deduplication across the whole list.

tomasstrba (approved):
LGMT! 👏 Just couple of comments below, but no blockers

Sources/Suggestions/Score.swift
brindy: Yeah that example won't work now because:
* It won't match the first character any more
* The second character wouldn't match anyway

But honestly, I think that's gonna be rare?  I assume people are more likely to search for `HBO` than `"` or `"HBO`

I can make it an OR condition though if you feel strongly?

tomasstrba: The new solution looks better to me 👍 Thank you, Brindy!

Issue #1019, opened 2024-10-15
Blocked requests inside nested "about:blank" iFrames are not counted properly


alisha:
We found a minor bug in the DuckDuckGo macOS browser (version 1.102.0) where some requests are blocked, but are not counted in DuckDuckGo's count of how many trackers have been blocked. This occurs when requests are made inside nested "about:blank" iFrames.

## Reproducing

1. Visit a website; we test on rocketnews24.com
2. Open a new tab, and observe how many things are blocked (see screenshot below)

![DDG Trackers Blocked - Original](https://github.com/user-attachments/assets/2fbef4e5-a9f5-4243-a5de-87127c9b213f)

3. Create a pair of nested frames with source "about:blank." We do this by entering the following code in the JavaScript console.

```
const [pframe, cframe] = (_ => {const parentFrame = document.createElement('iframe');const childFrame = document.createElement('iframe');parentFrame.name = "/parent-frame";parentFrame.id = "parent-frame";childFrame.name = "/child-frame";childFrame.id = "child-frame";document.body.appendChild(parentFrame);parentFrame.contentDocument.body.appendChild(childFrame);return  [parentFrame, childFrame];})()
```

4. Add a script that DDG will block to the parent frame. We do this by entering the following code in the JavaScript console.

```
((aSrc, frameElm) => {const scriptElm = document.createElement('script'); scriptElm.src = aSrc; frameElm.contentDocument.body.appendChild(scriptElm);})('https://rtb.openx.net/openrtbb/prebidjs', pframe);
```

5. Open a new tab, and observe how many things are blocked (see screenshot below). We find that the count increases by one, which is the expected behavior.

![DDG Trackers Blocked - Parent Frame](https://github.com/user-attachments/assets/b309b58b-69cc-4f0e-a499-db20fe1e5838)

6. Add the same script to the child frame. We do this by entering the following code in the JavaScript console.

```
((aSrc, frameElm) => {const scriptElm = document.createElement('script'); scriptElm.src = aSrc; frameElm.contentDocument.body.appendChild(scriptElm);})('https://rtb.openx.net/openrtbb/prebidjs', cframe);
```
7. Open a new tab, and observe how many things are blocked (see screenshot below). The count has not increased, but we expect the count to increase.

![DDG Trackers Blocked - Child Frame](https://github.com/user-attachments/assets/ab6f34f9-68f4-4aa7-92ff-2ebbb314527b)

## Impact

[...]

PR #1020, opened 2024-10-15
Fix how blocked requests made inside nested 'about:blank' iFrames are counted (issue #1019)

alisha:
This PR fixes https://github.com/duckduckgo/BrowserServicesKit/issues/1019; we confirmed this fixes the issue on the DuckDuckGo macOS browser (version 1.102.0) on macOS 14.5


jonathanKingston:
I think we should be looking at tab URL and not parent, that's the bigger issue here.

I think we probably want to copy this code instead:  https://github.com/duckduckgo/content-scope-scripts/blob/655d12d9e77f93cc36875142aa2ff3e06c52608d/injected/src/utils.js#L135-L157

jonathanKingston:
Closing this out in favour of https://github.com/duckduckgo/BrowserServicesKit/pull/1021

Whilst this fixed the specific case of about:blank frames we still needed to fix all other grand child frames not matching against the tab URL and instead the parent frame.

Again thanks for raising this issue and making us aware of the problem itself.

PR #1032, opened 2024-10-22
Bug Fix: Phishing Detection Dataset Discrepancies

not-a-rootkit:

**Required**:

Task/Issue URL: https://app.asana.com/0/1204023833050360/1208567121137949/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3469
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3440

**Optional**:

Tech Design URL:
CC:

**Description**:
In [Implement desktop integration efficacy tests - 5-7 days](https://app.asana.com/0/1207943168535188/1207205745934704/f) it was discovered that Swift's client-side caching results in out-of-date datasets and significant dataset discrepancies between different clients. For example, it's very common for the same request to return different results from the backend, resulting in a client believing they are updating to a newer revision than they are. Over time, this compounds and results in disparate versions of the same dataset across different clients, putting users at risk of landing on newer phishing pages.

Fix: 
 - Remove Client Side Caching in PhishingDetectionClient.swift
 - Ensure embedded dataset is used to replace the on-disk dataset when the revision of the embedded dataset > on disk dataset

**Steps to test this PR**:
1. Check unit tests
3. Change on-disk revision:
4. `echo "1650000" > "/System/Volumes/Data/Users/<user>/Library/Application Support/com.duckduckgo.macos.browser.debug/revision.txt"`
5. Build the browser
6. Visit https://privacy-test-pages.site/security/badware/phishing.html
7. Ensure blocked
8. Check on-disk revision:
9. `cat "/System/Volumes/Data/Users/<user>/Library/Application Support/com.duckduckgo.macos.browser.debug/revision.txt"`
10. Should be > 1650000

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Tests/PhishingDetectionTests/PhishingDetectionDataStoreTests.swift
+++ b/Tests/PhishingDetectionTests/PhishingDetectionDataStoreTests.swift
@@ -87,7 +87,7 @@ class PhishingDetectionDataStoreTests: XCTestCase {
         XCTAssertEqual(actualHashPrefix, embeddedHashPrefix)
         XCTAssertEqual(actualRevision, 5)
     }
-    
+
     func testWhenEmbeddedRevisionOlderThanOnDisk_ThenDontLoadEmbedded() async {
         let encoder = JSONEncoder()
         // On Disk Data Setup
@@ -186,7 +186,7 @@ class PhishingDetectionDataStoreTests: XCTestCase {
         let decoder = JSONDecoder()
         if let storedFilterSet = try? decoder.decode(Set<Filter>.self, from: storedFilterSetData!),
            let storedHashPrefixes = try? decoder.decode(Set<String>.self, from: storedHashPrefixesData!) {
-            
+
             XCTAssertEqual(storedFilterSet, updatedFilterSet)
             XCTAssertEqual(storedHashPrefixes, updatedHashPrefixes)
         } else {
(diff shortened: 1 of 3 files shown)


Tests/PhishingDetectionTests/PhishingDetectionDataStoreTests.swift
```diff
@@ -61,6 +61,22 @@ class PhishingDetectionDataStoreTests: XCTestCase {
         XCTAssertEqual(actualHashPrefix, expectedHashPrefix)
     }
 
+    func testWhenEmbeddedRevisionNewerThanOnDisk_ThenLoadEmbedded() async {
```
SabrinaTardio, line 64: You should set the dataStore sets and revision to something else as well  and then check which one of the two wins…
At the moment if loadRevisionFromDisk() < dataProvider.embeddedRevision { this line of code is not called at all from the test.

Also I would recommend to check both cases disk higher and embedded higher

SabrinaTardio (commented):
Steps work as described

I think we should keep the caching and if dos not work as it should means we are getting wrong values from the server… we should therefore update the server instead.

SabrinaTardio (approved):
LGTM!
There are some linting issues

PR #1033, opened 2024-10-25
Subscription oauth v2

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1207991044706235/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3480
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3580
What kind of version bump will this require?: **Major**
CC: @miasma13 

iOS PR: https://github.com/duckduckgo/iOS/pull/3480
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3580

**Description**:

This PR introduces the use of OAuth V2 authentication in Privacy Pro Subscription.
The code changes are comprehensive due to the paradigm changes between the old access token lifecycle and the new JWT lifecycle.
The Subscription UI and UX should be unchanged.

**Steps to test this PR**:
Test all Privacy Pro Subscription features and UX, more details [here](https://app.asana.com/0/1205842942115003/1208723559486361/f)

---
###### Internal references:


Sources/MaliciousSiteProtection/API/APIClient.swift
```diff
@@ -29,6 +29,7 @@ extension APIClient {
 extension APIClient: APIClient.Mockable {}
 
 public protocol APIClientEnvironment {
+    func queryItems(for requestType: APIRequestType) -> QueryItems
```
federicocappelli, line 32: Adopting the improved Networking V2

Sources/Common/CodableHelper.swift
```diff
@@ -0,0 +1,55 @@
...
+            encoder.outputFormatting = .prettyPrinted
+            return try encoder.encode(object)
+        } catch let error {
+            Logger.general.error("Error encoding input: \(error.localizedDescription, privacy: .public)")
+        }
+        return nil
+    }
+}
+
+public typealias DecodableHelper = CodableHelper
```
miasma13, line 55: Is this required?

federicocappelli: Yes but just because the branch is long-living, 90% of the "dumb" merge conflicts I get are for this name change, I'll remove it after we merge this.

Sources/Subscription/Flows/Models/SubscriptionOptions.swift
```diff
@@ -17,27 +17,41 @@
...
+        let name: SubscriptionEntitlement
+    }
+
     let platform: SubscriptionPlatformName
     let options: [SubscriptionOption]
-    let features: [SubscriptionFeature]
+    /// The available features in the subscription based on the country and feature flags. Not based on user entitlements
+    let features: [SubscriptionOptions.Feature]
+
+    public init(platform: SubscriptionPlatformName, options: [SubscriptionOption], availableEntitlements: [SubscriptionEntitlement]) {
```
miasma13, line 32: What is the reason for changing the naming here from `features` to `availableEntitlements`. This is confusing and against the naming conventions we agreed upon: 
- entitlements are permissions granted to user's account for accessing features (e.g. account has entitlement  to use VPN)
- features are services bundled together as a subscription plan, set of features defines the complete set of entitlements user is granted when subscribing

https://app.asana.com/0/1208524871249522/1208561749459707/f
https://app.asana.com/0/1208524871249522/1208561749459707/f

Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -171,20 +282,166 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
             }
         } else {
-            return [.networkProtection, .dataBrokerProtection, .identityTheftRestoration]
+            let result = [SubscriptionFeature(entitlement: .networkProtection, enabled: true),
+                          SubscriptionFeature(entitlement: .dataBrokerProtection, enabled: true),
+                          SubscriptionFeature(entitlement: .identityTheftRestoration, enabled: true)]
+            Logger.subscription.debug("Default Subscription features: \(result)")
+            return result
+        }
+    }
```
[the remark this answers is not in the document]
miasma13: Also what is worth, this method is to be used by client apps to determine which parts of the UI are to be shown for the currently active subscription.

Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -171,20 +282,166 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
+
+    public func isFeatureActive(_ entitlement: SubscriptionEntitlement) async -> Bool {
+        let currentFeatures = await currentSubscriptionFeatures(forceRefresh: false)
+        return currentFeatures.contains { feature in
+            feature.entitlement == entitlement && feature.enabled
```
miasma13, line 444: Similar as comment above. I strongly would encourage to stick to naming conventions already established and rename it to `hasEntitlement(_:)`. Also there should be no logic crosschecking it with the entitlements here, just simple check for entitlements array if one is present and return true or false.
I think I see what you tried to achieve here but the whole quite simple concept becomes convoluted without reason. `Features` in the subscription define what UI-wise is part of current subscription, and `Entitlements` define which parts user currently has access to within the scope of current subscription feature set.

Sources/Subscription/Flows/Models/SubscriptionOptions.swift
federicocappelli: The reasons are the following: 

The new source of truth for both user entitlements and subscription entitlements is a single function in SubscriptionManager, this function needs to return "something" that explains what features are available and if they are available for the specific user, this something natural name is Feature (entitlement + enabled or not), I tried to invent something different but the result was confusing. 
How we obtain entitlements and features has changed a lot, getting stuck on the previous naming convention didn't make much sense.

The data we get from the backend is identical between user entitlement and subscription entitlement, they are all entitlement so I maintained the same approach, the moment we merge the two concept we get a SubscriptionFeature

The old Feature and Entitlement struct with `var product` was verbose for no reason, the only thingh we care of is the entitlement

The important concept here is that all of these entitlements and where they come from are internal implementation details, the only externally visible object is
```
public struct SubscriptionFeature: Equatable, CustomDebugStringConvertible {
    public var entitlement: SubscriptionEntitlement
    public var enabled: Bool
    ...
}
```

Sources/Subscription/Managers/SubscriptionManager.swift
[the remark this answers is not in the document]
federicocappelli: | A subscription is defined as being a bundle of features
yes, in the backend. We enhance the "subscription features" by adding additional info (available or not to the user)

I'm not sure where's the problem here, everything you described is right and it's how is currently implemented, in fact, the number/type of features returned by this function is dictated by the entitlement (backend name) present in the subscription and they are marked enabled/disabled by the user capabilities.

As you say this function user is the client's UI that needs a way to decide which features displaying (here is the name) and if they are disabled or not :) Exactly what this function provides.

Note that SubscriptionManager (in this function in particular) is the only entity that has the right to manipulate entitlements and create Features, nobody else touches them so the joining of user entitlements and subscription entitlements needs to happen here.

I think the misunderstanding here is that you are not separating the internal implementation from the Subscription framework interface. The models' name and functions' contract should be dictated by the need of the user (in this case the App's UI) not from some internal implementation ideas.

Sources/Subscription/Managers/SubscriptionManager.swift
federicocappelli: As explained above `hasEntitlement(_:)` doesn't make sense anymore, nobody outside SubscriptionManager can or should know/care about internal implementation details (The S of SOLID principles) so it's the duty of `currentSubscriptionFeatures` to provide the right information needed from the UI, we can't ask the users (main apps) to check if the use entitlement is present, get the subscription entitlements, compare them and then update the UI.

THISISDINOSAUR:
There's basically no PIR stuff in BSK at the moment

Sources/Common/UserDefaultsCache.swift
```diff
@@ -44,7 +44,7 @@ public class UserDefaultsCache<ObjectType: Codable> {
         let object: ObjectType
     }
 
-    let logger = { Logger(subsystem: Bundle.main.bundleIdentifier ?? "DuckDuckGo", category: "UserDefaultsCache") }()
+    let logger = { Logger(subsystem: "UserDefaultsCache", category: "") }()
```
diegoreymendez, line 47: This will remove these entries from VPN log export.  Is there a need to make this change?

Sources/NetworkProtection/Status/ControlllerErrorMessageObserver/ControllerErrorMesssageObserverThroughDistributedNotifications.swift
```diff
@@ -61,6 +61,8 @@ public class ControllerErrorMesssageObserverThroughDistributedNotifications: Con
         let errorMessage = notification.object as? String
         logErrorChanged(isShowingError: errorMessage != nil)
 
+        Logger.networkProtectionStatusReporter.debug("Received error message")
```
diegoreymendez, line 64: The message is not correct.  If `errorMessage == nil` then there's no error message.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -515,15 +516,15 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+    open func load(options: StartupOptions) async throws {
         loadKeyValidity(from: options)
         loadSelectedEnvironment(from: options)
         loadSelectedServer(from: options)
         loadSelectedLocation(from: options)
         loadDNSSettings(from: options)
         loadTesterEnabled(from: options)
 #if os(macOS)
-        try loadAuthToken(from: options)
+            try await loadAuthToken(from: options)
```
diegoreymendez, line 527: The indentation is off here as this is not a conditional block, but a conditionally-compiled line.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -677,11 +680,8 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
         self.snoozeTimingStore.reset()
 
         do {
-            try load(options: startupOptions)
-
-            if (try? tokenStore.fetchToken()) == nil {
-                throw TunnelError.startingTunnelWithoutAuthToken
-            }
+            try await load(options: startupOptions)
+            Logger.networkProtection.log("Startup options loaded correctly")
```
diegoreymendez, line 684: The error handling needs to be improved here.  Previously we could only have `startingTunnelWithoutAuthToken`, but now we can get a number of different possible errors from `tokenProvider.adopt` and `tokenProvider.getTokenContainer`.

These errors should ideally implement `CustomNSError` and include underlying error information.  See `PacketTunnelProvider.TunnelError` for reference.

The reason this is important is because it's what'll allow us to debug tunnel start issues coming from these changes, and understand what's going on.

diegoreymendez (commented):
Adding some comments.  Still going through the code.

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -123,7 +123,7 @@ final class NetworkProtectionConnectionTester {
     }
 
     func stop() {
-        Logger.networkProtectionConnectionTester.log("🔴 Stopping connection tester")
+        Logger.networkProtectionConnectionTester.log("🟢 Stopping connection tester")
```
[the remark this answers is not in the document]
federicocappelli: oops, bed copy-paste, fixed

Sources/Common/KeychainType.swift
```diff
@@ -46,6 +44,8 @@ public enum KeychainType {
             }
         case .system:
             return [kSecUseDataProtectionKeychain: false]
+        case .fileBased:
```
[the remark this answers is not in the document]
federicocappelli: Good point, KeychainType was duplicated across BSK and the main apps, I just merged the two (or 3?) implementation and left all cases for backward compatibility. Removed now across all apps.

Sources/NetworkProtection/Status/ControlllerErrorMessageObserver/ControllerErrorMesssageObserverThroughDistributedNotifications.swift
federicocappelli: log removed completely, it was a debugging leftover

Sources/NetworkProtection/PacketTunnelProvider.swift
federicocappelli: Understood, atm getTokenContainer can fail for a variety of reasons, from storage to API, decoding and decryption errors, we don't have a single set Error so implementing this is quite difficult, should I wrap any possible error into a higher level TunnelError?

diegoreymendez:
Note: I've edited the description to add links to the related PRs for easy of navigation.

Sources/Common/UserDefaultsCache.swift
diegoreymendez: Let's leave it like this, I think this may not affect how we're retrieving the logs, actually.

That said, in the context of such a large PR this is the type of changes that I'd prefer not be mixed in, as they add noise and aren't strictly necessary for a long lived & large PR

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -189,7 +188,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
                           selectionMethod: NetworkProtectionServerSelectionMethod) async throws -> (server: NetworkProtectionServer,
                                                                                                     newExpiration: Date?) {
 
-        guard let token = try? tokenStore.fetchToken() else { throw NetworkProtectionError.noAuthTokenFound }
+        let token = try await VPNAuthTokenBuilder.getVPNAuthToken(from: tokenProvider, policy: .localValid)
```
diegoreymendez, line 191: Could you wrap this in a `NetworkProtectionError` with underlying error data?

It'll make it much easier to debug issues if we know where this error is coming from.

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
```diff
@@ -102,9 +103,7 @@ public actor NetworkProtectionDeviceManager: NetworkProtectionDeviceManagement {
     /// This method will return the remote server list if available, or the local server list if there was a problem with the service call.
     ///
     public func refreshServerList() async throws -> [NetworkProtectionServer] {
-        guard let token = try? tokenStore.fetchToken() else {
-            throw NetworkProtectionError.noAuthTokenFound
-        }
+        let token = try await VPNAuthTokenBuilder.getVPNAuthToken(from: tokenProvider, policy: .localValid)
```
diegoreymendez, line 106: Could you wrap this in a `NetworkProtectionError` with underlying error data?

It'll make it much easier to debug issues if we know where this error is coming from.

Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -104,70 +196,229 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
+            if (try await oAuthClient.migrateV1Token()) != nil {
+                pixelHandler(.v1MigrationSuccessful)
 
-            defer {
-                completion(isSubscriptionActive)
+                // cleaning up old data
+                clearSubscriptionCache()
             }
+        } catch {
+            Logger.subscription.error("Failed to migrate V1 token: \(error, privacy: .public)")
```
diegoreymendez, line 221: Shouldn't this error throw?  The VPN calls this at startup: what happens if the migration fails?

Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -104,70 +196,229 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
+        // Fetching fresh subscription
+        if isUserAuthenticated {
+            do {
+                let subscription = try await getSubscription(cachePolicy: .reloadIgnoringLocalCacheData)
+                Logger.subscription.log("Subscription is \(subscription.isActive ? "active" : "not active", privacy: .public)")
+                if subscription.isActive {
+                    pixelHandler(.subscriptionIsActive)
                 }
+            } catch {
+                Logger.subscription.error("Failed to load initial subscription data: \(error, privacy: .public)")
```
diegoreymendez, line 234: Shouldn't this error throw? The VPN calls this at startup: what happens if the user is not authenticated?  Can they still connect?

diegoreymendez (commented):
Added some additional comments, still going through all this.

THISISDINOSAUR (commented):
There's almost nothing PIR related in BSK so I have no comments here

Sources/Common/KeychainType.swift
```diff
@@ -1,7 +1,7 @@
 //
 //  KeychainType.swift
 //
-//  Copyright © 2023 DuckDuckGo. All rights reserved.
+//  Copyright © 2024 DuckDuckGo. All rights reserved.
```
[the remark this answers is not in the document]
federicocappelli: KeychainType.swift existed in both bsk and mac app, I just moved it, reverting it back just to avoid noise

Sources/Subscription/Managers/SubscriptionManager.swift
federicocappelli: If the migration fails there's nothing we can do, we just don't have a token and we stop and just send a pixel (added now in the vpn too)

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
federicocappelli: I have modified Networking.OAuthClientError.missingTokens by adding an underlying error and used in here

Sources/NetworkProtection/NetworkProtectionDeviceManager.swift
federicocappelli: I have modified Networking.OAuthClientError.missingTokens by adding an underlying error and used in here

Sources/NetworkProtection/KeyManagement/NetworkProtectionKeychainStore.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 import Common
 import os.log
 
-enum NetworkProtectionKeychainStoreError: Error, NetworkProtectionErrorConvertible {
+public enum NetworkProtectionKeychainStoreError: Error, NetworkProtectionErrorConvertible {
```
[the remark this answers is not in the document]
federicocappelli: correct, reverted

Sources/NetworkProtection/Diagnostics/NetworkProtectionConnectionTester.swift
```diff
@@ -123,7 +123,7 @@ final class NetworkProtectionConnectionTester {
     }
 
     func stop() {
-        Logger.networkProtectionConnectionTester.log("🔴 Stopping connection tester")
+        Logger.networkProtectionConnectionTester.log("⚫️ Stopping connection tester")
```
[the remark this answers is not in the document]
federicocappelli: all the other stopping colours are black, this is the only one red and confusing, but I reverted it

miasma13 (commented):
First batch of comments

Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -104,70 +196,231 @@ public final class DefaultSubscriptionManager: SubscriptionManager {
...
         }
     }
+
+    public func isFeatureActive(_ entitlement: SubscriptionEntitlement) async -> Bool {
+        guard isUserAuthenticated else { return false }
+
+        let currentFeatures = await currentSubscriptionFeatures(forceRefresh: false)
+        return currentFeatures.contains { feature in
+            feature.entitlement == entitlement && feature.enabled
+        }
```
miasma13, line 424: I find this naming confusing. Let's go one by one, `currentSubscriptionFeatures` above is clear and well defined, features that are provided by current subscription. But when you move to `isFeatureActive` and `SubscriptionFeature.enabled` it can be interpreted in multiple ways (e.g. `isFeatureActive(.networkProtection)`, does it mean that VPN is running? etc.).
- `isFeatureActive` - means if the feature is provided within the current subscription and if current user has entitlements to use it, previously we sticked to established vocabulary and checked if used has entitlements to a feature. If you don't like this maybe something along the line `isFeatureAvailableForUser`?
- `SubscriptionFeature.enabled` - this is the underlying property storing Bool for the above so I would align them together

Sources/Subscription/API/SubscriptionEndpointService.swift
```diff
@@ -63,118 +71,189 @@ public protocol SubscriptionEndpointService {
...
 
-    // MARK: -
-
-    public func getProducts() async -> Result<[GetProductsItem], APIServiceError> {
-        await apiService.executeAPICall(method: "GET", endpoint: "products", headers: nil, body: nil)
+    public func clearSubscription() {
+        cacheSerialQueue.sync {
+            subscriptionCache.reset()
+        }
+//        NotificationCenter.default.post(name: .subscriptionDidChange, object: self, userInfo: nil)
```
miasma13, line 194: Please double check if this is needed.

miasma13 (commented):
Second part of the review

Sources/Networking/OAuth/OAuthClient.swift
```diff
@@ -0,0 +1,433 @@
...
+    public func removeLocalAccount() {
+        Logger.OAuthClient.log("Removing local account")
+        tokenStorage.tokenContainer = nil
+        legacyTokenStorage?.token = nil
+    }
+
+    /* MARK: Edit account
+
+    /// Helper, single use
+    public class AccountEditor {
```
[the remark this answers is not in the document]
federicocappelli: Agree, all removed.

Sources/Networking/OAuth/SessionDelegate.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+
+import Foundation
+import os.log
+
+public final class SessionDelegate: NSObject, URLSessionTaskDelegate {
+
+    /// Disable automatic redirection, in our specific OAuth implementation we manage the redirection, not the user
+    public func urlSession(_ session: URLSession, task: URLSessionTask, willPerformHTTPRedirection response: HTTPURLResponse, newRequest request: URLRequest) async -> URLRequest? {
+        return nil
```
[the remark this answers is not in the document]
federicocappelli: Do you mean the entire `SessionDelegate`? no, it's used in the OAuthClient init in the main apps
```
let urlSession = URLSession(configuration: configuration,
                                    delegate: SessionDelegate(),
                                    delegateQueue: nil)
```

Sources/Subscription/Managers/SubscriptionManager.swift
```diff
@@ -18,49 +18,141 @@
...
+    public var entitlement: SubscriptionEntitlement
+    public var enabled: Bool
+
+    public var debugDescription: String {
+        "\(entitlement.rawValue) is \(enabled ? "enabled" : "disabled")"
+    }
+}
+
+/// The sole entity responsible of obtaining, storing and refreshing an OAuth Token
+public protocol SubscriptionTokenProvider {
```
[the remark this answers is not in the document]
federicocappelli: I moved SubscriptionFeature and SubscriptionTokenProvider to dedicated files

Sources/Subscription/Managers/SubscriptionManager.swift
federicocappelli: I agree:
- `isFeatureActive` renamed `isFeatureAvailableForUser`
- `SubscriptionFeature.enabled ` renamed `availableForUser`

Sources/Subscription/API/SubscriptionEndpointService.swift
federicocappelli: Not needed for now, anything related to the subscription cancellation is handled by the signOut that sends the proper notification

Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -19,159 +19,209 @@
...
+                Logger.subscriptionAppStorePurchaseFlow.log("Failed to restore an account from a past purchase: \(error.localizedDescription, privacy: .public)")
+                do {
+                    externalID = try await subscriptionManager.getTokenContainer(policy: .createIfNeeded).decodedAccessToken.externalID
+                } catch OAuthClientError.deadToken {
+                    do {
+                        let transactionJWS = try await recoverSubscriptionFromDeadToken()
+                        return .success(transactionJWS)
+                    } catch {
+                        return .failure(.purchaseFailed(OAuthClientError.deadToken))
                     }
```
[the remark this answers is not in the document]
federicocappelli: `restoreAccountFromPastPurchase  fails because the token is dead, the first action of `recoverSubscriptionFromDeadToken` is to sign out the user (aka deleting everything, including the dead token) and re-run `restoreAccountFromPastPurchase`. The assumption is that the user had a token (dead) so a subscription was present, so the same subscription is recoverable with a new token.

Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -19,159 +19,209 @@
...
-                return true
-            case .failure:
-                return false
+        Logger.subscriptionAppStorePurchaseFlow.log("Completing Subscription Purchase")
+
+        subscriptionManager.clearSubscriptionCache()
+
+        do {
+            let subscription = try await subscriptionManager.confirmPurchase(signature: transactionJWS)
+            if subscription.isActive {
```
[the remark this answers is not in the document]
federicocappelli: I'm not sure how to handle this but:
- The request has a 3 max retries with a 4s delay between retries `APIRequestV2.RetryPolicy(maxRetries: 3, delay: 4.0)` but the call itself doesn't fail if the subscription is inactive
- If the subscription is inactive we log it and return a failure to the script page with the correct error, so I would like to detect it still. I have never seen a failure here, I'm just trying to fail gracefully in case of BE errors.
- I moved the token refresh out of the if so it is done no matter what.

Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -19,159 +19,209 @@
...
-        let subscriptionInfo = await subscriptionEndpointService.getSubscription(accessToken: token, cachePolicy: .reloadIgnoringLocalCacheData)
-
-        // Only return an externalID if the subscription is expired
-        // To prevent creating multiple subscriptions in the same account
-        if case .success(let subscription) = subscriptionInfo,
-           !subscription.isActive,
-            subscription.platform != .apple {
-            return externalID
+    @discardableResult
+    private func recoverSubscriptionFromDeadToken() async throws -> String {
```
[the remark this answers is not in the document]
federicocappelli: I agree, this is confusing, so I did the following:
- removed the recovery function call from every `AppStorePurchaseFlow` call and returned the dead token error normally 
- made recoverSubscriptionFromDeadToken() public and generic, so can be used from averyone receiving a dead token error, I'll update the clients accordingly

Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -19,159 +19,209 @@
...
+        Logger.subscriptionAppStorePurchaseFlow.log("Recovering Subscription From Dead Token")
+
+        // Clear everything, the token is unrecoverable
+        await subscriptionManager.signOut(notifyUI: true)
```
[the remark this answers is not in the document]
federicocappelli: not really, we want to log out the user if every recovery attempt fails and we want to notify the UI accordingly

Sources/Subscription/Flows/AppStore/AppStorePurchaseFlow.swift
```diff
@@ -19,159 +19,209 @@
...
         // Make the purchase
         switch await storePurchaseManager.purchaseSubscription(with: subscriptionIdentifier, externalID: externalID) {
         case .success(let transactionJWS):
             return .success(transactionJWS)
         case .failure(let error):
-            Logger.subscription.error("[AppStorePurchaseFlow] purchaseSubscription error: \(String(reflecting: error), privacy: .public)")
-            accountManager.signOut(skipNotification: true)
+            Logger.subscriptionAppStorePurchaseFlow.error("purchaseSubscription error: \(String(reflecting: error), privacy: .public)")
+
+            await subscriptionManager.signOut(notifyUI: true)
```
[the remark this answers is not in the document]
federicocappelli: The old behaviour is very different from the current one, I re-tested this and notifying the Ui at logout here is needed in case the user has multiple windows and the subscription settings app is visible

Sources/Subscription/API/Model/PrivacyProSubscription.swift
```diff
@@ -64,4 +66,37 @@ public struct Subscription: Codable, Equatable {
...
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateStyle = .medium
+        dateFormatter.timeStyle = .short
+        dateFormatter.timeZone = TimeZone.current
+        return dateFormatter.string(from: date)
+    }
+
+//    public static func == (lhs: PrivacyProSubscription, rhs: PrivacyProSubscription) -> Bool {
+//        return lhs.productId == rhs.productId &&
+//        lhs.name == rhs.name &&
```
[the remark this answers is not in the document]
federicocappelli: deleted

PR #1052, opened 2024-10-31
Ignore GRDB

Bunn:
Task/Issue URL:  https://app.asana.com/0/1204167627774280/1208651137074643/f
iOS PR: N/A
macOS PR: N/A
What kind of version bump will this require?: N/A

**Description**:
Ignore GRDB

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,3 @@
-# Please see the documentation for all configuration options:
-# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
 version: 2
 updates:
   - package-ecosystem: "gitsubmodule" # See documentation for possible values
@@ -11,3 +8,5 @@ updates:
     directory: "/" # Location of Package.swift file
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "GRDB"


.github/dependabot.yml
```diff
@@ -10,4 +7,6 @@ updates:
   - package-ecosystem: "swift" # For Swift package updates
     directory: "/" # Location of Package.swift file
     schedule:
-      interval: "daily"
\ No newline at end of file
+      interval: "daily"
+    ignore:
+      - dependency-name: "GRDB"
```
ayoy, line 12: @Bunn do you know if this is a package name or repo name? Because while package name is `GRDB`, repo name is `GRDB.swift` (https://github.com/duckduckgo/GRDB.swift)

ayoy (approved):
LGTM but just one comment to clarify the dependency name. Fine to merge it and see if it works and then correct as needed :) Thanks a lot @Bunn!

.github/dependabot.yml
Bunn: Honestly, I'm not sure, the [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) are not very specific: `Use to ignore updates for dependencies with matching names`

Since it only says "name" and not "repo name" I'm assuming it's just `GRDB` 
If it doesn't work we can update to `GRDB*` or `GRDB.swift`

PR #1054, opened 2024-11-01
Validate VPN errors before re-throwing them

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/414709148257752/1208225499545869/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3513
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3490

**Description**:

This PR attempts to catch malformed errors before they get thrown to the OS. It also informs us when this happens.

**Steps to test this PR**:

To test this, we need to make sure that existing errors still throw as expected, and that errors we're trying to filter out get successfully caught.

**For testing existing errors on iOS:**

1. Open PacketTunnelProvider.swift
2. Go to `func startTunnel(options: [String: NSObject]? = nil) async throws`, which is the function where errors from the rest of the VPN get thrown to the OS
3. Pick one of the `try` functions in `startTunnel`, for example `try await startTunnel(onDemand: startupOptions.startupMethod == .automaticOnDemand)`
4. Inside that function, throw an error such as `TunnelError.simulateTunnelFailureError`
5. Now, attempt to start the VPN, and check that it fails
6. After it has failed, go to the iOS debug menu's `Network Protection` section, and check the metadata at the very bottom to confirm that the `lastDisconnectError` value to make sure it matches the error you used in step 4

**For testing errors that we're filtering out:**

First, add the following error somewhere in PacketTunnelProvider:

```
    enum ErrorThatCrashes: CustomNSError {
        case crash

        var errorCode: Int {
            return 123
        }

        static var errorDomain: String {
            return "error-domain"
        }

        var errorUserInfo: [String: Any] {
            return [
                NSUnderlyingErrorKey: "testing"
            ]
        }
    }
```

This error uses an underlying error that isn't a valid error object, which causes a VPN extension crash. Next:

1. Throw an instance of that error from the same place that you used in the first set of instructions
2. Check the Debug menu's Network Protection section and confirm that you see a valid error object in the metadata which is of the type `InvalidDiagnosticError`
3. Also, check that it contains an underlying error that is just domain and code

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -738,7 +738,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             self.knownFailureStore.lastKnownFailure = KnownFailure(error)
 
             providerEvents.fire(.tunnelStartAttempt(.failure(error)))
-            
+
             // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
             throw validated(error: error)
         }
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -771,8 +771,6 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
     private func startTunnel(onDemand: Bool) async throws {
         do {
-            throw ErrorThatCrashes.crash
-
             Logger.networkProtection.log("Generating tunnel config")
             Logger.networkProtection.log("Excluded ranges are: \(String(describing: self.settings.excludedRanges), privacy: .public)")
             Logger.networkProtection.log("Server selection method: \(self.currentServerSelectionMethod.debugDescription, privacy: .public)")
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -692,9 +692,6 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                 throw TunnelError.startingTunnelWithoutAuthToken
             }
         } catch {
-            // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
-            let error = validated(error: error)
-
             if startupOptions.startupMethod == .automaticOnDemand {
                 // If the VPN was started by on-demand without the basic prerequisites for
                 // it to work we skip firing pixels.  This should only be possible if the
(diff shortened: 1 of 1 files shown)


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -1815,6 +1821,56 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
+            case .errorWithInvalidUnderlyingError(let error):
+                return "Error '\(type(of: error))', message: \(error.localizedDescription)"
+            }
+        }
+
+        var errorUserInfo: [String: Any] {
+            switch self {
+            case .errorWithInvalidUnderlyingError(let error):
+                let newError = NSError(domain: (error as NSError).domain, code: (error as NSError).code)
+                return [NSUnderlyingErrorKey: newError]
```
samsymons, line 1847: We want to know what error violated the underlying error validation, but without actually including its underlying errors - knowing its domain and code should be enough to debug further.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -692,6 +692,9 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                 throw TunnelError.startingTunnelWithoutAuthToken
             }
         } catch {
+            // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
+            let error = validated(error: error)
```
diegoreymendez, line 696: This effectively removes the underlying error info from `error`.  I imagine the idea here is to catch this in the pixel we fire in line 710, but I'm hesitant to discard valuable info from that pixel.

Would you be open to:
- Moving this down to line 715, so that pixels aren't affected; and
- Optionally adding a new temporary pixel specifically to register these cases?

The downside is it's a bit more work to push a new pixel out, but the upside is we'll be able to match this pixel to the failures we're already getting, including all of the underlying error chain.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -723,6 +726,9 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
 
             providerEvents.fire(.tunnelStartAttempt(.success))
         } catch {
+            // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
+            let error = validated(error: error)
```
diegoreymendez, line 730: Same suggestion as above.

diegoreymendez (commented):
Added a comment.

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: I've updated this to only validate the error right as it's being thrown, such that the pixel call isn't affected.

I deliberately sent it with the pixel for starters so that we'd get a signal that this new error type occurred, and it would attach the original error as its underlying error, but I agree we can define something new for that instead.

diegoreymendez (approved):

PR #1081, opened 2024-11-15
add experiment logic

SabrinaTardio:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1204186595873227/1208687542689225/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3582
macOS PR:  https://github.com/duckduckgo/macos-browser/pull/3559

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1208680035673275/f
CC:

**Description**: updates stateFor and isSubfeatureEnabled api to manage remote experiments and also adds getActiveExperimentsAPI

**Steps to test this PR**:
1. Check all tests pass
2. Run macOS app
3.  Run iOS app

<!—
Before submitting a PR, please ensure you have tested the combinations you expect the reviewer to test, then delete configurations you *know* do not need explicit testing.

Using a simulator where a physical device is unavailable is acceptable.
—>

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

—
###### Internal references:


Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -210,15 +266,66 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+    private func checkCohortState(_ subfeatureData: PrivacyConfigurationData.PrivacyFeature.Feature,
+                                  passedCohort: CohortID?,
+                                  assignCohortEnabled: Bool,
+                                  subfeatureID: SubfeatureID,
+                                  parentFeature: PrivacyFeature) -> PrivacyConfigurationFeatureState {
+        let cohorts = subfeatureData.cohorts ?? []
+        let targetsState = checkTargets(subfeatureData)
+        let assignIfEnabled = assignCohortEnabled && targetsState == .enabled
+        let assignedCohortResponse = experimentManager.cohort(for: ExperimentSubfeature(parentID: parentFeature.rawValue, subfeatureID: subfeatureID, cohorts: cohorts), assignIfEnabled: assignIfEnabled)
+        let possibleDisabledReason: PrivacyConfigurationFeatureDisabledReason = assignedCohortResponse.didAttemptAssignment &&  targetsState != .enabled ? .targetDoesNotMatch : .experimentCohortDoesNotMatch
```
jaceklyp, line 306: unnecessary space before &&

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -180,28 +188,76 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
+                        assignCohortEnabled: false,
+                        versionProvider: versionProvider,
+                        randomizer: randomizer)
+    }
+
+    public func stateFor(_ subfeature: any PrivacySubfeature,
+                         cohortID: CohortID?,
+                         versionProvider: AppVersionProvider,
+                         randomizer: (Range<Double>) -> Double) -> PrivacyConfigurationFeatureState {
+        Self.experimentManagerQueue.sync {
```
jaceklyp, line 231: You should limit the sync queue block to only what truly requires synchronization. Avoid including any dependencies inside it (!). For example, the InternalUserDecider should handle its responsibilities elsewhere because if someone adds main queue synchronization code to it ever, it will lead to potential deadlock

Sources/BrowserServicesKit/PrivacyConfig/AppPrivacyConfiguration.swift
```diff
@@ -210,15 +266,66 @@ public struct AppPrivacyConfiguration: PrivacyConfiguration {
...
         return .enabled
     }
 
+    private func matchTargets(targets: [PrivacyConfigurationData.PrivacyFeature.Feature.Target]) -> Bool {
+        return targets.contains { target in
+            (target.localeCountry == nil || target.localeCountry == locale.regionCode) &&
+            (target.localeLanguage == nil || target.localeLanguage == locale.languageCode)
+        }
+    }
+
```
jaceklyp, line 328: What I dislike the most is that AppPrivacyConfiguration is turning into a MassiveAppPrivacyConfiguration. Can we plan a follow-up project to streamline and clean this up, possibly?

jaceklyp (changes requested):

bwaresiak:
Just glancing over this, made me thing that we should add @samsymons  and @diegoreymendez as stakeholders, as Config space is now used both in Host app and the Extensions - so potentially we could be using Feature experiments in all of these targets, while current implementation doesn't really solve the potential problem of cross-process data synchronization. 

I'm not saying we need to solve for this now, but rather should be aware of that any any potential limitations.

Other than this, I see some good discussions in the comments, @SabrinaTardio and @jaceklyp :-) If you would feel like I could help speed up things by jumping on a call to try talking through some concepts brought here with both you, feel free to ping me. Other than that I was thinking to have a look (overview, not detailed) at this (as it's pretty interesting problem!), but once you are closer to actual solution to not add noise to your current flow.

SabrinaTardio:
> Just glancing over this, made me think that we should add @samsymons and @diegoreymendez as stakeholders, as Config space is now used both in Host app and the Extensions - so potentially we could be using Feature experiments in all of these targets, while current implementation doesn't really solve the potential problem of cross-process data synchronization.
> 
> I'm not saying we need to solve for this now, but rather should be aware of that any any potential limitations.
> 
> Other than this, I see some good discussions in the comments, @SabrinaTardio and @jaceklyp :-) If you would feel like I could help speed up things by jumping on a call to try talking through some concepts brought here with both you, feel free to ping me. Other than that I was thinking to have a look (overview, not detailed) at this (as it's pretty interesting problem!), but once you are closer to actual solution to not add noise to your current flow.

Just for transparency we had a sync discussion about this and have a TD as a result (https://app.asana.com/0/1204186595873227/1208823859841917/f)
I’m working on either updating this or make an entirely new one for simplicity.

SabrinaTardio:
We had a sync discussion about this and have a TD as a result (https://app.asana.com/0/1204186595873227/1208823859841917/f)
I will open a new PR with the changes

PR #1091, opened 2024-11-22
Malware protection 1: rename PhishingDetection to MaliciousSiteProtection

mallexxx:

**Required**:

Task/Issue URL: https://app.asana.com/0/1202406491309510/1208033567421351/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3642
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3588

**Description**:
- Renamed `PhishingDetection` to `MaliciousSiteProtection`
- Adjusted and cleaned up some code parts 

**Steps to test this PR**:
1. Activate Feature Flag override for malicious site protections in Debug -> Feature Flag overrides
2. Visit https://privacy-test-pages.site/security/badware/phishing.html, validate phishing detection works without changes


Sources/MaliciousSiteProtection/MaliciousSiteDetector.swift
```diff
@@ -104,27 +83,29 @@ public class PhishingDetector: PhishingDetecting {
...
-        let hashPrefix = generateHashPrefix(for: canonicalHost, length: hashPrefixStoreLength)
-        if dataStore.hashPrefixes.contains(hashPrefix) {
-            // Check local filterSet first
-            if checkLocalFilters(canonicalHost: canonicalHost, canonicalUrl: canonicalUrl) {
-                return true
-            }
-            // If nothing found, hit the API to get matches
-            if await checkApiMatches(canonicalHost: canonicalHost, canonicalUrl: canonicalUrl) {
-                return true
+    public func evaluate(_ url: URL) async -> ThreatKind? {
```
not-a-rootkit, line 92: I think this is fine for now, but soon this will result in two API requests when we introduce malware protections. It also doesn't allow for independent feature flagging. We should discuss if/how we want to do this on Monday in our call. Doesn't block this PR though.

Sources/MaliciousSiteProtection/Services/EmbeddedDataProvider.swift
```diff
@@ -58,8 +58,7 @@ public class PhishingDetectionDataProvider: PhishingDetectionDataProviding {
              let filterSetData = try loadData(from: embeddedFilterSetURL, expectedSHA: embeddedFilterSetDataSHA)
              return try JSONDecoder().decode(Set<Filter>.self, from: filterSetData)
          } catch {
-             Logger.phishingDetectionDataProvider.error("🔴 Error: SHA mismatch for filterSet JSON file. Expected \(self.embeddedFilterSetDataSHA)")
-             return []
+             fatalError("🔴 Error: SHA mismatch for filterSet JSON file. Expected \(self.embeddedFilterSetDataSHA)")
```
not-a-rootkit, line 61: I know we do this for privacy configs, but I'm not a fan of crashing the app in this instance. However, ideally this would never make it past internal testing so I'm okay with it.

Sources/MaliciousSiteProtection/Services/FileStore.swift
```diff
@@ -0,0 +1,68 @@
...
+    func read(from filename: String) -> Data?
+}
+
+public struct FileStore: FileStoring {
+    private let dataStoreURL: URL
+
+    public init() {
+        let dataStoreDirectory: URL
+        do {
+            dataStoreDirectory = try FileManager.default.url(for: .applicationSupportDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
```
not-a-rootkit, line 33: Again, doesn't affect this PR but just making notes for myself: in the next PR we should try to align with @alessandroboron's TD for storage locations: https://app.asana.com/0/481882893211075/1207273224076495/f

not-a-rootkit (approved):
Just a few notes for future reference but nothing blocking this PR. Tests seem to be working and end-to-end experience seems unaffected.

Sources/MaliciousSiteProtection/Services/EmbeddedDataProvider.swift
mallexxx: Reaching the `catch` block means the app bundle corruption (same is valid for a missing Storyboard), and considering the [Let it crash](https://app.asana.com/0/1199230911884351/1199550018748895/f) discussion this makes impossible to run the app further and we must crash here.
Plus, it happens early in the app lifetime during smoke testing that makes it easily noticeable and not affecting the production built.

Sources/MaliciousSiteProtection/Services/FileStore.swift
mallexxx: this part is not modified from the original implementation: to be update in later PRs to match the config files storage location

Sources/MaliciousSiteProtection/MaliciousSiteDetector.swift
mallexxx: good, let‘s mark this as a discussion point, I was assuming the API would check for both threat kinds and return the detected kind if any;

PR #1092, opened 2024-11-25
Malware protection 2: refactor APIClient

mallexxx:

**Required**:

Task/Issue URL: https://app.asana.com/0/481882893211075/1208033567421351/f
iOS PR: 
macOS PR: 

**Optional**:

Tech Design URL:
CC:

**Description**:
- Refactor `APIClient` to use `Networking.APIRequestV2`; implement generic request/response
- Fix query argument percent-encoding issue in APIRequestV2

**Steps to test this PR**:
1. Validate Phishing Protection change sets and Match API calls work as before

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/MaliciousSiteProtection/API/APIClient.swift
```diff
@@ -18,129 +18,119 @@
...
-                return response
-            } else {
-                logDebug("🔸 Failed to decode response for \(String(describing: responseType)): \(data)")
-            }
-        } catch {
-            logDebug("🔴 Failed to load \(String(describing: responseType)) data: \(error)")
-        }
-        return nil
+    public func matches(forHashPrefix hashPrefix: String) async throws -> APIClient.Response.Matches {
+        let result = try await load(.matches(hashPrefix: hashPrefix))
```
not-a-rootkit, line 133: Can we pass a shorter timeout to the /matches endpoint? This could block navigation so we should favour navigation loading if the backend is degraded. On Android we're looking at a maximum 1 second timeout for this request. The other requests can be heavier but they are background tasks anyway so not as risky.

Sources/MaliciousSiteProtection/API/APIClient.swift
```diff
@@ -18,129 +18,119 @@
...
-        let queryItems = [URLQueryItem(name: "hashPrefix", value: hashPrefix)]
-        guard let url = createURL(for: .matches, queryItems: queryItems) else {
-            logDebug("🔸 Invalid matches URL: \(hashPrefix)")
-            return []
-        }
-        return await fetch(url: url, responseType: MatchResponse.self)?.matches ?? []
+    public func load<Request: APIRequestProtocol>(_ requestConfig: Request) async throws -> Request.ResponseType {
+        let requestType = requestConfig.requestType
+        let headers = environment.headers(for: requestType)
+        let url = environment.url(for: requestType)
```
not-a-rootkit, line 109: Are there any specific client caching configs we should consider here? The server should return a TTL of 10 minutes for the API so I imagine this will be automatically respected here, but just checking.

not-a-rootkit (approved):
LGTM. Just left a comment on cache policies and timeouts. I've noticed that the current production deployment has a very long cache policy of 1 year, which causes some issues right now in internal. 

I don't think it's worth changing the cache policy on the client though, as I'm working on a task to fix the cache headers on the backend separately. Might be good to just add a stricter timeout on /matches calls.

Sources/MaliciousSiteProtection/API/APIClient.swift
mallexxx: I‘d say no as the caching policies should be set on the backend side: if there‘s caching it should be fine as all of our requests are parametrized (i.e. revision, hash etc)

Sources/MaliciousSiteProtection/API/APIClient.swift
mallexxx: Added parametrized Timeout

PR #1097, opened 2024-11-26
Add Privacy Stats module for collecting stats about blocked trackers

ayoy:

**Required**:

Task/Issue URL: https://app.asana.com/0/72649045549333/1208246350498754/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3651
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3611

**Optional**:

Tech Design URL: https://app.asana.com/0/481882893211075/1208848285586302
CC: @bwaresiak 

**Description**:
This change adds PrivacyStats module that will be used in macOS HTML New Tab Page.
Privacy Stats uses Core Data for storage where it keeps blocked trackers counts per tracker
company name per day, for past 7 days.
PrivacyStats class is the main (and only) public interface. It provides simple API to fetch current
stats, clear them (for Fire button use) and record blocked trackers, as well as a publisher
that notifies about changes to stats.

**Steps to test this PR**:
TBD in the macOS PR

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/Common/Extensions/DateExtension.swift
```diff
@@ -54,6 +54,10 @@ public extension Date {
         return Calendar.current.isDate(date1, inSameDayAs: date2)
     }
 
+    static func isSameHour(_ date1: Date, _ date2: Date) -> Bool {
+        return Calendar.current.isDate(date1, equalTo: date2, toGranularity: .hour)
```
aataraxiaa, line 58:
```suggestion
        Calendar.current.isDate(date1, equalTo: date2, toGranularity: .hour)
```

Sources/Common/Extensions/DateExtension.swift
```diff
@@ -64,7 +68,11 @@ public extension Date {
     }
 
     var startOfDay: Date {
-        return  Calendar.current.startOfDay(for: self)
+        return Calendar.current.startOfDay(for: self)
+    }
+
+    func daysAgo(_ days: Int) -> Date {
+        return Calendar.current.date(byAdding: .day, value: -days, to: self)!
```
aataraxiaa, line 75:
```suggestion
        Calendar.current.date(byAdding: .day, value: -days, to: self)!
```

Sources/Common/Extensions/DateExtension.swift
aataraxiaa, line 58:
```suggestion
        Calendar.current.isDate(date1, equalTo: date2, toGranularity: .hour)
```

Sources/PrivacyStats/PrivacyStats.swift
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * This protocol describes `PrivacyStats` interface.
+ */
+public protocol PrivacyStatsCollecting {
+
+    /**
+     * Record a tracker for a given `companyName`.
+     *
+     * `PrivacyStats` implementation calls the actor under the hood,
```
aataraxiaa, line 77: To ease understanding think we could say here:

```suggestion
     * `PrivacyStats` implementation calls the `CurrentPack` actor under the hood,
```

Sources/PrivacyStats/PrivacyStats.swift
```diff
@@ -0,0 +1,268 @@
...
+                try context.save()
+            } catch {
+                Logger.privacyStats.error("Faild to load current stats: \(error)")
+                errorEvents?.fire(.failedToLoadCurrentPrivacyStats(error))
+            }
+        }
+        return pack ?? PrivacyStatsPack(timestamp: Date.currentPrivacyStatsPackTimestamp)
+    }
+
+    private func subscribeToAppTermination() {
```
aataraxiaa, line 248: I understand why it’s a nice approach to encapsulate this login here, as the client does not need to worry about reacting to `willTerminate` events. However, two thoughts:

1. I’m unsure if the `PrivacyStats` “layer" should be aware of app events (and importing `AppKit` and `UIKit`). 
2. The same way other `PrivacyStats` actions are explicit public methods (e.g recording a tracker), maybe committing changes should be an explicit action, which the client needs to be conscious of. At the moment it feels like a side effect which is happening, hidden from the clients

So, one suggestion would be to make this a public function which clients need to call in their own `willTerminate` methods.

Sources/PrivacyStats/internal/CurrentPack.swift
```diff
@@ -0,0 +1,116 @@
...
+import Foundation
+import os.log
+
+/**
+ * This actor provides thread-safe access to an instance of `PrivacyStatsPack`.
+ *
+ * It's used by `PrivacyStats` class to record blocked trackers that can possibly
+ * come from multiple open tabs (web views) at the same time.
+ */
+actor CurrentPack {
```
aataraxiaa, line 29: Nice `actor` implementation 👍🏼

Sources/PrivacyStats/internal/DailyBlockedTrackersEntity.swift
```diff
@@ -0,0 +1,52 @@
...
+
+@objc(DailyBlockedTrackersEntity)
+final class DailyBlockedTrackersEntity: NSManagedObject {
+
+    @nonobjc class func fetchRequest() -> NSFetchRequest<DailyBlockedTrackersEntity> {
+        return NSFetchRequest<DailyBlockedTrackersEntity>(entityName: "DailyBlockedTrackersEntity")
+    }
+
+    class func entity(in context: NSManagedObjectContext) -> NSEntityDescription {
+        return NSEntityDescription.entity(forEntityName: "DailyBlockedTrackersEntity", in: context)!
```
aataraxiaa, line 30: Minor: As String here is used twice could extract to a private static constant.

Sources/PrivacyStats/internal/DailyBlockedTrackersEntity.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+
+import Common
+import CoreData
+
+@objc(DailyBlockedTrackersEntity)
+final class DailyBlockedTrackersEntity: NSManagedObject {
+
+    @nonobjc class func fetchRequest() -> NSFetchRequest<DailyBlockedTrackersEntity> {
+        return NSFetchRequest<DailyBlockedTrackersEntity>(entityName: "DailyBlockedTrackersEntity")
```
aataraxiaa, line 26:
```suggestion
        NSFetchRequest<DailyBlockedTrackersEntity>(entityName: "DailyBlockedTrackersEntity")
```

Sources/PrivacyStats/internal/DailyBlockedTrackersEntity.swift
aataraxiaa, line 30:
```suggestion
        NSEntityDescription.entity(forEntityName: "DailyBlockedTrackersEntity", in: context)!
```

aataraxiaa: Also, are we confident we can force unwrap here?

Sources/PrivacyStats/internal/PrivacyStatsUtils.swift
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Deletes stats older than 7 days for all companies.
+     */
+    static func deleteOutdatedPacks(in context: NSManagedObjectContext) {
+        let oldestValidTimestamp = Date().privacyStatsOldestPackTimestamp
+
+        let request = DailyBlockedTrackersEntity.fetchRequest()
+        request.predicate = NSPredicate(format: "%K < %@", #keyPath(DailyBlockedTrackersEntity.timestamp), oldestValidTimestamp as NSDate)
+        context.deleteAll(matching: request)
```
aataraxiaa, line 109: Eventually this calls a `NSManagedObjectContext` extension which loops over entities and deletes them all. Would it be more efficient to use a `NSBatchDeleteRequest` here? e.g:

```
let request = NSBatchDeleteRequest(fetchRequest: DailyBlockedTrackersEntity.fetchRequest())
request.predicate = NSPredicate(format: "%K < %@", Keys.timestamp, oldestValidTimestamp as NSDate)
try context.execute(request)
```

aataraxiaa (commented):
Left a few comments @ayoy. Overall looks good. Nice tests. 👍🏼

Sources/Common/Extensions/DateExtension.swift
ayoy: I'm actually going to delete it because it's not used anywhere :) thanks for spotting.

Sources/PrivacyStats/PrivacyStats.swift
ayoy: I appreciate you questioning this and I agree it feels off. The reason I included it here was that HistoryTabExtension does a similar thing but that one is fully in the macOS browser. Let me move it away from BSK and only leave the function here to allow saving the state on demand. I'll also remove the RunLoop thingy (also copied over from another place in the app) because this isn't some crucial data that we're storing here :)

Sources/PrivacyStats/internal/PrivacyStatsUtils.swift
ayoy: Great idea, tbf I forgot NSBatchDeleteRequest existed :) updated it now and verified with tests that it works well!

Sources/PrivacyStats/internal/DailyBlockedTrackersEntity.swift
ayoy: Yes, we do it for all entity descriptions across BSK.

Sources/PrivacyStats/PrivacyStats.swift
ayoy: I need to keep the RunLoop around anyway because it's an asynchronous function. But just so you know, it's a common pattern used in the app and looks safe to me.

aataraxiaa (approved):
Thanks for making the changes @ayoy , looks great!

Sources/PrivacyStats/internal/PrivacyStatsUtils.swift
```diff
@@ -0,0 +1,121 @@
...
+     * Deletes stats older than 7 days for all companies.
+     */
+    static func deleteOutdatedPacks(in context: NSManagedObjectContext) throws {
+        let oldestValidTimestamp = Date().privacyStatsOldestPackTimestamp
+
+        let fetchRequest = NSFetchRequest<NSFetchRequestResult>(entityName: DailyBlockedTrackersEntity.Const.entityName)
+        fetchRequest.predicate = NSPredicate(format: "%K < %@", #keyPath(DailyBlockedTrackersEntity.timestamp), oldestValidTimestamp as NSDate)
+        let deleteRequest = NSBatchDeleteRequest(fetchRequest: fetchRequest)
+
+        try context.execute(deleteRequest)
```
bwaresiak, line 111: As we are using batch delete request, I'd reset the context after this to invalidate any object that could be still cached in the object graph by the context.

Sources/PrivacyStats/internal/CurrentPack.swift
```diff
@@ -0,0 +1,116 @@
...
+        commitTask?.cancel()
+    }
+
+    /**
+     * This function is used when clearing app data, to clear any stats cached in memory.
+     *
+     * It sets a new empty pack with the current timestamp.
+     */
+    func resetPack() {
+        resetStats(andSet: Date.currentPrivacyStatsPackTimestamp)
```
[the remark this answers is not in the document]
ayoy: Great point - it actually should. It works because it's used by the Fire button which closes everything and opens a new window in which case new data is pulled, but overall this should trigger a notification. Thanks for spotting it!

Sources/PrivacyStats/PrivacyStats.swift
```diff
@@ -0,0 +1,253 @@
...
+
+public final class PrivacyStats: PrivacyStatsCollecting {
+
+    public static let bundle = Bundle.module
+
+    public let statsUpdatePublisher: AnyPublisher<Void, Never>
+
+    private let db: CoreDataDatabase
+    private let context: NSManagedObjectContext
+    private var currentPack: CurrentPack?
```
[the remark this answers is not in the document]
bwaresiak: You could make it static and pass context and errorEvents as params to avoid this warning.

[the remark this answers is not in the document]
ayoy: That was my initial idea actually. Agreed that making it a `let` has a benefit to the reader so let me update it 👍

PR #1115, opened 2024-12-05
Fix threading issue between using a Semaphore and async/await

bwaresiak:

**Required**:

Task/Issue URL: https://app.asana.com/0/856498667320406/1208887821349124/f
iOS PR: tba
macOS PR:tba 

**Description**:

Fix threading issue between using a Semaphore and async/await

**Steps to test this PR**:

Ensure clean install runs fine (no lookup).
Ensure consecutive runs load well (with lookup).
Smoke test startups.

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/BrowserServicesKit/ContentBlocking/ContentBlockingRulesLookupTask.swift
+++ b/Sources/BrowserServicesKit/ContentBlocking/ContentBlockingRulesLookupTask.swift
@@ -28,31 +28,45 @@ extension ContentBlockerRulesManager {
 
         private let sourceManagers: [ContentBlockerRulesSourceManager]
 
-        public private(set) var result: [LookupResult]?
-
         init(sourceManagers: [ContentBlockerRulesSourceManager]) {
             self.sourceManagers = sourceManagers
         }
 
-        func lookupCachedRulesLists() async throws {
+        func lookupCachedRulesLists() throws -> [LookupResult]  {
+
+            let models = sourceManagers.compactMap { $0.makeModel() }
+            if models.count != sourceManagers.count {
+                // We should only load rule lists, in case we can match every one of the expected ones
+                throw WKError(.contentRuleListStoreLookUpFailed)
+            }
 
             var result = [LookupResult]()
-            for sourceManager in sourceManagers {
-                guard let model = sourceManager.makeModel() else {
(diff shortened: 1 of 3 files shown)


Sources/BrowserServicesKit/ContentBlocking/ContentBlockingRulesLookupTask.swift
```diff
@@ -28,31 +28,45 @@ extension ContentBlockerRulesManager {
...
+                        group.leave()
+                    }
                 }
+            }
+
+            let operationResult = group.wait(timeout: .now() + 6)
 
-                result.append((ruleList, model))
+            guard operationResult == .success, result.count == models.count else {
+                throw WKError(.contentRuleListStoreLookUpFailed)
```
samsymons, line 66: Do we currently get enough signal that this issue is happening?

I'd want to know how often this specific case happens to users as it provides some interesting information about the `lookUpContentRuleList` API, though to the caller we don't differentiate between reasons why this function failed. From what I see, a timeout is indistinguishable from a failed `models.count != sourceManagers.count` check above, for example.

bwaresiak: I think this should not happen, unless WebKit has latency, which is not even actionable by us.

We actually have different signal added recently by Shilpa: startup distribution of lookup vs last fetched vs recompile flows. I'm hesitant to ad anything here without good reason, and it's not like timeout affects app state in some really bad way - will just require us to recompile (which does a lookup anyway).

The checks for models are onl for the app upgrades really, when we change the set of Content Blocking data sets (e.g. when introducing CTL or Attribution features).

That's why I don't really think we need the Pixels here, what do you think?

samsymons: > We actually have different signal added recently by Shilpa: startup distribution of lookup vs last fetched vs recompile flows

Great point! As long as we have a way to get a hunch when this case could be happening (and I agree it shouldn't), I'm happy here. Thanks!

samsymons (approved):

PR #1116, opened 2024-12-05
Apple client send and receive support for crash report cohort IDs (CRCIDs)

jdjackson:

**Required**:

Task/Issue URL: https://app.asana.com/0/1208592102886666/1208759541597499/f
iOS & macOS PRs: I’m looking for feedback on these BSK changes before posting app-level PRs
What kind of version bump will this require?: Minor (is this right? 😊).
CC: @larryturtis

**Optional**:
Tech Design URL: https://app.asana.com/0/1208592102886666/1208660326715650/f

**Description**:
DO NOT MERGE - this is a draft for input, not ready to go live yet.  Before this is ready to merge:
- I’ll have to coordinate iOS and macOS changes to match, as the API signatures for the CrashCollection class have changed
- I will be waiting past the December 9 code freeze, so these changes are not included in the last public release of 2024, and instead have an extended internal testing period

**Steps to test this PR**:
1. Relying on included automated tests is the primary testing function, but this can be tested live against @larryturtis’s dev instance.  To do this, set CrashReportSender line 35 to:
`static let reportServiceUrl = URL(string: "https://9e3c-20-75-144-152.ngrok-free.app/crash.js")!`

I have tested via the following means:
- Unit/integration tests included in this PR.
- Local changes to the iOS app to integrate with these BSK changes, seeing crashes submit to @larryturtis’s dev instance, which in turn routes crash through to [our Sentry instance (development environment)](https://errors.duckduckgo.com/organizations/ddg/issues/?environment=development&project=8&statsPeriod=90d), and ensuring that the client sends, receives, and updates, its crcid when appropriate.

Note that I have _not_ been able to test equivalent changes on macOS because I’m unable to add my computer to the DDG provisioning profile (see https://app.asana.com/0/1200194497630846/1208709596010744/f).

**OS Testing**:

* [ ] iOS 14
* [X] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

—
###### Internal references:

### DIFF
--- a/Sources/Crashes/CrashCollection.swift
+++ b/Sources/Crashes/CrashCollection.swift
@@ -101,10 +101,11 @@ public final class CrashCollection {
                                     Logger.general.debug("Received matching value for CRCID: \(receivedCRCID), no update necessary")
                                 }
                             } else {
-                                Logger.general.debug("No value for CRCID header: \(Const.crcidKey), clearing local crcid value")
+                                Logger.general.debug("No value for CRCID header: \(Const.crcidKey), clearing local crcid value if present")
                                 self.crashCollectionStorage.removeObject(forKey: Const.crcidKey)
                             }
                         case .failure(let failure):
+                            // TODO: Is it worth sending a pixel for this case, so that we can monitor for missing crash reports?
                             Logger.general.debug("Crash Collection - Sending Crash Report: failed (\(failure))")
                         }
                     }
--- a/Sources/Crashes/CrashCollection.swift
+++ b/Sources/Crashes/CrashCollection.swift
@@ -18,6 +18,7 @@
 
 import Foundation
 import MetricKit
+import Persistence
 
 public enum CrashCollectionPlatform {
     case iOS, macOS, macOSAppStore
@@ -38,9 +39,10 @@ public enum CrashCollectionPlatform {
 @available(iOS 13, macOS 12, *)
 public final class CrashCollection {
 
-    public init(platform: CrashCollectionPlatform) {
-        crashHandler = CrashHandler()
-        crashSender = CrashReportSender(platform: platform)
+    public init(crashReportSender: CrashReportSending, crashCollectionStorage: KeyValueStoring = UserDefaults()) {
+        self.crashHandler = CrashHandler()
+        self.crashSender = crashReportSender
+        self.crashCollectionStorage = crashCollectionStorage
     }
 
     public func start(didFindCrashReports: @escaping (_ pixelParameters: [[String: String]], _ payloads: [Data], _ uploadReports: @escaping () -> Void) -> Void) {
@@ -173,16 +175,17 @@ public final class CrashCollection {
 
     var isFirstCrash: Bool {
         get {
-            UserDefaults().object(forKey: Const.firstCrashKey) as? Bool ?? true
+            crashCollectionStorage.object(forKey: Const.firstCrashKey) as? Bool ?? true
         }
 
         set {
-            UserDefaults().set(newValue, forKey: Const.firstCrashKey)
+            crashCollectionStorage.set(newValue, forKey: Const.firstCrashKey)
         }
     }
 
     let crashHandler: CrashHandler
-    let crashSender: CrashReportSender
+    let crashSender: CrashReportSending
+    let crashCollectionStorage: KeyValueStoring
 
     enum Const {
         static let firstCrashKey = "CrashCollection.first"
--- a/Sources/Crashes/CrashCollection.swift
+++ b/Sources/Crashes/CrashCollection.swift
@@ -19,6 +19,7 @@
 import Foundation
 import MetricKit
 import Persistence
+import os.log
 
 public enum CrashCollectionPlatform {
     case iOS, macOS, macOSAppStore
@@ -88,13 +89,18 @@ public final class CrashCollection {
                     for payload in processedData {
                         let crcid = self.crashCollectionStorage.object(forKey: Const.crcidKey) as? String
                         let result =  await self.crashSender.send(payload, crcid: crcid)
-                        // TODO: Logic for handling response here
-                        switch result {
-                        case .success(let success):
-                            let receivedCRCID = "faked" // Pull from data (decode json)
-                            self.crashCollectionStorage.set(receivedCRCID, forKey: Const.crcidKey)
+
+                        switch result.result {
+                        case .success:
+                            Logger.general.debug("Successfully sent crash report")
+                            if let receivedCRCID = result.response?.allHeaderFields[Const.crcidHTTPHeaderKey] as? String {
+                                self.crashCollectionStorage.set(receivedCRCID, forKey: Const.crcidKey)
+                            } else {
+                                Logger.general.debug("No value found for header: \(Const.crcidKey), clearing local crcid value")
+                                self.crashCollectionStorage.removeObject(forKey: Const.crcidKey)
+                            }
                         case .failure(let failure):
-                            // TODO: ?
+                            Logger.general.debug("Failed to send crash report: \(failure)")
                         }
                     }
                     didFinishHandlingResponse()
@@ -204,5 +210,6 @@ public final class CrashCollection {
     enum Const {
         static let firstCrashKey = "CrashCollection.first"
         static let crcidKey = "CrashCollection.crcid"
+        static let crcidHTTPHeaderKey = "x-crcid"   // TODO: Settle on final value
     }
 }
--- a/Sources/Crashes/CrashCollection.swift
+++ b/Sources/Crashes/CrashCollection.swift
@@ -57,7 +57,11 @@ public final class CrashCollection {
     ///   - didFindCrashReports: callback called after payload preprocessing is finished.
     ///     Provides processed JSON data to be presented to the user and Pixel parameters to fire a crash Pixel.
     ///     `uploadReports` callback is used when the user accepts uploading the crash report and starts crash upload to the server.
-    public func start(process: @escaping ([MXDiagnosticPayload]) -> [Data], didFindCrashReports: @escaping (_ pixelParameters: [[String: String]], _ payloads: [Data], _ uploadReports: @escaping () -> Void) -> Void) {
+    public func start(process: @escaping ([MXDiagnosticPayload]) -> [Data],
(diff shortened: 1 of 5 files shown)


Sources/Crashes/CrashCollection.swift
```diff
@@ -80,8 +87,29 @@ public final class CrashCollection {
...
+                                    self.crashCollectionStorage.set(receivedCRCID, forKey: Const.crcidKey)
+                                } else {
+                                    Logger.general.debug("Received matching value for CRCID: \(receivedCRCID), no update necessary")
+                                }
+                            } else {
+                                Logger.general.debug("No value for CRCID header: \(Const.crcidKey), clearing local crcid value if present")
+                                self.crashCollectionStorage.removeObject(forKey: Const.crcidKey)
+                            }
+                        case .failure(let failure):
+                            // TODO: Is it worth sending a pixel for this case, so that we can monitor for missing crash reports?
```
samsymons, line 108: I think we should do this - we've had crashes go missing quietly when Apple moved to the IPS format on macOS, which was only detected after noticing how quiet the crash reports had been lately.

Sources/Crashes/CrashReportSender.swift
```diff
@@ -19,28 +19,86 @@
...
+    case invalidResponse
+}
+
+// By conforming to a protocol, we can sub in mocks more easily
+public final class CrashReportSender: CrashReportSending {
 
     static let reportServiceUrl = URL(string: "https://duckduckgo.com/crash.js")!
+
+    static let httpHeaderCRCID = "crcid"
+
```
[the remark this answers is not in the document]
jdjackson: Interesting.  I see the swiftlint errors now that the PR checks have run, but no mention of which files/lines are actually causing issues (it’s just a bunch of "Error: Lines should not have trailing whitespace (trailing_whitespace)” and similar at the end of the SwiftLint output here in github anyway.  Is there some way to see swiftlint output with file/line references I’m missing?

Sources/Crashes/CrashCollection.swift
jdjackson: Makes sense, I called it out because I thought it might be good idea but didn’t want to hold up starting to get feedback on my changes so far while I dug further.  Where should I look for a getting started guide / instructions / steps to follow for defining new pixels?  Presumably somewhere in Asana, but… asana search :(

Sources/Crashes/CrashCollection.swift
```diff
@@ -80,8 +87,29 @@ public final class CrashCollection {
...
                     for payload in processedData {
-                        await self.crashSender.send(payload)
+                        let crcid = self.crashCollectionStorage.object(forKey: Const.crcidKey) as? String
+                        let result =  await self.crashSender.send(payload, crcid: crcid)
+
+                        switch result.result {
+                        case .success:
+                            Logger.general.debug("Crash Collection - Sending Crash Report: succeeded")
+                            if let receivedCRCID = result.response?.allHeaderFields[CrashReportSender.httpHeaderCRCID] as? String {
+                                if crcid != receivedCRCID {
```
LarryTurtis, line 97: why not just always set the received crcid, regardless of whether it matches or not? Are there performance implications?

Sources/Crashes/CrashCollection.swift
samsymons: Yep, we do - check [this task](https://app.asana.com/0/69071770703008/1201077660132463/f) out. It'll tell you how to define a new pixel, which will likely lead you out to the Privacy Triage pixel definition form that'll ask you a range of questions.

Sources/Crashes/CrashCollection.swift
jdjackson: It’s unlikely to be an issue either way, but in my experience it’s best practice not to re-store an existing value, because it can trigger code observing a value when the value hasn’t actually changed.

Sources/Crashes/CrashReportSender.swift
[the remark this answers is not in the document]
jdjackson: SwiftLint is still complaining, but I’m not seeing why/where, and I’m not inclined to spend a lot of time on this.  Let’s file (as we have) under “swiftlint should work right” once we have a monorepo, cool? :)

jdjackson:
Ok, I’ve updated BSK again so we can support crashes in the macOS app as well.  To do this, I’ve created a new CRCIDManager class that is responsible for crcid storage and handling logic around updating, or not, the CRCID based on server response.  This lets the macOS client (non-appstore) use the same logic without relying on CrashCollection, which is only used by macOS app store builds.

Sources/Crashes/CrashCollection.swift
jdjackson: Pixel support added, but I’m not sure how to test it beyond watching client side logs.  Are there dashboards somewhere that show all our pixels?  Do I need to add monitoring for these new pixels somehow?

Sources/Crashes/CrashReportSender.swift
```diff
@@ -18,29 +18,88 @@
...
-        do {
-            _ = try await session.data(for: request)
-        } catch {
-            assertionFailure("CrashReportSender: Failed to send the crash report")
+        Logger.general.debug("CrashReportSender: Awaiting session data")
+        let task = session.dataTask(with: request) { data, response, error in
+            if let response = response as? HTTPURLResponse {
+                Logger.general.debug("CrashReportSender: Received HTTP response code: \(response.statusCode)")
+                if response.statusCode == 200 {
+                    response.allHeaderFields.forEach { print("\($0.key): \($0.value)") }    // TODO: Why do we straight-up print these, rather than debug logging?
```
jdjackson, line 69: @samsymons @loremattei @ayoy Does it make sense for me to convert this to a `Logger.general.debug` call instead of a simple `print`?

Sources/Crashes/CrashCollection.swift
```diff
@@ -38,9 +41,12 @@ public enum CrashCollectionPlatform {
 @available(iOS 13, macOS 12, *)
 public final class CrashCollection {
 
-    public init(platform: CrashCollectionPlatform) {
-        crashHandler = CrashHandler()
-        crashSender = CrashReportSender(platform: platform)
+    public init(crashReportSender: CrashReportSending,
+                crashCollectionStorage: KeyValueStoring = UserDefaults()) {
```
samsymons, line 45: Let's swap this out with the `standard` instance:

```suggestion
                crashCollectionStorage: KeyValueStoring = UserDefaults.standard) {
```

Sources/Crashes/CrashReportSender.swift
samsymons: @jdjackson Yep, definitely - that'll allow for more control over filtering etc. in the Xcode console.

Sources/Crashes/CrashReportSender.swift
```diff
@@ -18,29 +18,87 @@
...
-    public init(platform: CrashCollectionPlatform) {
+    private let session = URLSession(configuration: .ephemeral)
+
+    public init(platform: CrashCollectionPlatform, pixelEvents: EventMapping<CrashReportSenderError>?) {
         self.platform = platform
+        self.pixelEvents = pixelEvents
     }
 
-    public func send(_ crashReportData: Data) async {
+    public func send(_ crashReportData: Data, crcid: String?, completion: @escaping (_ result: Result<Data?, Error>, _ response: HTTPURLResponse?) -> Void) {
```
samsymons, line 52: The `crcid` argument here is optional, but the storage is returning a non-nil `String`, as it defaults to returning an empty string. I'd prefer to update the storage API (`public var crcid: String`) to return `String?` and avoid defaulting to an empty string, since it gives us type-safety around ensuring we have a _real_ CRC ID.

Sources/Crashes/CrashCollection.swift
```diff
@@ -80,8 +90,10 @@ public final class CrashCollection {
             didFindCrashReports(pixelParameters, processedData) {
                 Task {
                     for payload in processedData {
-                        await self.crashSender.send(payload)
+                        let result =  await self.crashSender.send(payload, crcid: self.crcidManager.crcid)
```
samsymons, line 93: `await`ing in a loop can take a long time if we have a lot of `payload` entries, but this code was already behaving this way so I'd recommend we leave it as-is for now.

Sources/Crashes/CrashCollection.swift
```diff
@@ -171,20 +183,72 @@ public final class CrashCollection {
...
+    let crashSender: CrashReportSending
+    let crashCollectionStorage: KeyValueStoring
+    let crcidManager: CRCIDManager
 
     enum Const {
         static let firstCrashKey = "CrashCollection.first"
     }
 }
+
+// TODO: This should really be its own file, but adding a new file to BSK and propagating it to iOS and macOS projects is hard.  This can be done as a separate PR once the main changes land across all 3 repos
```
samsymons, line 210: Let me know if you want to jump on a quick call and I can demo how to do this - I agree this should go into a new file and would be happy to help with that. 🙂

Sources/TestUtils/MockKeyValueStore.swift
```diff
@@ -19,11 +19,11 @@
 import Foundation
 import Persistence
 
-public class MockKeyValueStore: KeyValueStoring {
+public class MockKeyValueStore: NSObject, KeyValueStoring {
```
samsymons, line 22: We can undo the changes in this file since they were only added when a KVO expectation was being used in the test suite. Now that it's not, we can scrap the change.

Sources/Crashes/CrashCollection.swift
```diff
@@ -171,20 +183,72 @@ public final class CrashCollection {
...
+                crcid = ""
+            }
+        case .failure(let failure):
+            Logger.general.debug("Crash Collection - Sending Crash Report: failed (\(failure))")
+        }
+    }
+
+    public var crcid: String {
+        get {
+            return self.store.object(forKey: CRCIDManager.crcidKey) as? String ?? ""
```
samsymons, line 247: Here's where I'd suggest we update this to no longer default to an empty string in the getter, and instead do that defaulting at the call site of where we get the `crcid` - that'll help prevent us from being surprised one day when we expect to get a real value out of this getter, and instead we get an empty string.

samsymons (approved):

PR #1118, opened 2024-12-06
Update RMF version matching to ignore build number

samsymons:

**Required**:

Task/Issue URL: https://app.asana.com/0/1207619243206445/1208908529122506/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3686
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3635

**Description**:

This PR updates the RMF version matching logic to ignore build number. Previously, when you tried to match a version like `1.110.0`, it would fail because you needed to know the build number.

To fix this, this PR makes the following changes:
1. Update the version number fetching logic to use `versionNumber`, not `versionAndBuildNumber`
2. To avoid breaking messages that are using build number, the attributes that match version number ranges have been updated to trim the build number at init time (let's discuss that one in the comments though)

**Steps to test this PR**:
1. Check that the test suite is green and correctly covers this new case
1. Launch one of the clients and verify that existing messages are parsed correctly - many of the existing production messages contain version checks so these can be used for testing as-is, but please ping me if you'd prefer a custom message written to facilitate easier testing

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/RemoteMessaging/Model/MatchingAttributesPrototypes/StringRangeMatching.swift
+++ b/Sources/RemoteMessaging/Model/MatchingAttributesPrototypes/StringRangeMatching.swift
@@ -42,12 +42,7 @@ extension StringRangeMatching {
         let value = jsonMatchingAttribute[RuleAttributes.value] as? String ?? MatchingAttributeDefaults.stringDefaultValue
         let fallback = jsonMatchingAttribute[RuleAttributes.fallback] as? Bool
 
-        self.init(
-            min: min.trimmingBuildNumber,
-            max: max.trimmingBuildNumber,
-            value: value.trimmingBuildNumber,
-            fallback: fallback
-        )
+        self.init(min: min, max: max, value: value, fallback: fallback)
     }
 
     init(fallback: Bool?) {
@@ -65,18 +60,4 @@ extension StringRangeMatching {
         }
         return RangeStringNumericMatchingAttribute(min: min, max: max).matches(value: value)
     }
-}
-
-private extension String {
-
-    var trimmingBuildNumber: String {
-        let components = self.split(separator: ".")
-
-        if components.count == 4 {
-            return components.dropLast().joined(separator: ".")
-        } else {
-            return self
-        }
-    }
-
 }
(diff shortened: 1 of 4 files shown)


Tests/RemoteMessagingTests/Matchers/CommonAppAttributeMatcherTests.swift
```diff
@@ -26,6 +26,7 @@ import XCTest
 class CommonAppAttributeMatcherTests: XCTestCase {
 
     private var matcher: CommonAppAttributeMatcher!
+    private let versionNumber = "3.2.1"
```
samsymons, line 29: I've moved this to using a fixed version number in the test suite, since `AppVersion.shared.versionNumber` in macOS BSK tests returns the OS version, which is 2 components for me, which causes some problems in the tests.

Also, the approach of using a dynamic version number led to some weird test cases, like tests which decrement a component would produce version strings like `16.0.-1`.

Sources/RemoteMessaging/Model/MatchingAttributesPrototypes/StringRangeMatching.swift
```diff
@@ -42,7 +42,12 @@ extension StringRangeMatching {
         let value = jsonMatchingAttribute[RuleAttributes.value] as? String ?? MatchingAttributeDefaults.stringDefaultValue
         let fallback = jsonMatchingAttribute[RuleAttributes.fallback] as? Bool
 
-        self.init(min: min, max: max, value: value, fallback: fallback)
+        self.init(
+            min: min.trimmingBuildNumber,
+            max: max.trimmingBuildNumber,
+            value: value.trimmingBuildNumber,
```
samsymons, line 48: These lines exist to avoid breaking existing messages that already define a build number.

This approach means that build number trimming will happen for all `StringRangeMatching` attributes, which is currently `AppVersionMatchingAttribute` and `OSMatchingAttribute`. But, if you would prefer to keep this change limited to the app version attribute, I can override its initializer and push that logic down there instead.

Alternatively, it could be moved into `RangeStringNumericMatchingAttribute` instead. We have a few options, let me know what you prefer. Thanks!

ayoy: The logic we're applying here is specific to the app version attribute, right? I'd rather see it in the `AppVersionMatchingAttribute` to limit its reach. Unless you foresee this adjustment to be useful for other strings?

ayoy (approved):
LGTM, thanks for this @samsymons!

PR #1133, opened 2024-12-16
change api

SabrinaTardio:

**Required**:

Task/Issue URL:  https://app.asana.com/0/1204186595873227/1208964427775425/f
iOS PR: 
macOS PR: 

**Optional**:

Tech Design URL:
CC:

**Description**: Changes the name of the original fireExperimentPixel func and adds a new function where the framework does not manages the counting

**Steps to test this PR**:
1. All tests pass 

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:

### DIFF
--- a/Sources/PixelExperimentKit/PixelExperimentKit.swift
+++ b/Sources/PixelExperimentKit/PixelExperimentKit.swift
@@ -72,7 +72,7 @@ extension PixelKit {
         ExperimentConfig.fireFunction(event, .uniqueByNameAndParameters, false)
     }
 
-    /// Fires a pixel for a specific action in an experiment, based on conversion window and value thresholds (if value is a number).
+    /// Fires a pixel for a specific action in an experiment, based on conversion window and value.
     /// - Parameters:
     ///   - subfeatureID: Identifier for the subfeature associated with the experiment.
     ///   - metric: The name of the metric being tracked (e.g., "searches").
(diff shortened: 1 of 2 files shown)


Sources/PixelExperimentKit/PixelExperimentKit.swift
```diff
@@ -94,11 +94,43 @@ extension PixelKit {
...
         guard let experimentData = featureFlagger.getAllActiveExperiments()[subfeatureID] else { return }
 
+        // Define event
+        let event = event(for: subfeatureID, experimentData: experimentData, conversionWindowDays: conversionWindowDays, metric: metric, value: value)
+
+        // Send unique by name and parameter pixel if within conversion window
+        let isInWindow = isUserInConversionWindow(conversionWindowDays, enrollmentDate: experimentData.enrollmentDate)
+        if isInWindow {
+            ExperimentConfig.fireFunction(event, .uniqueByNameAndParameters, false)
+        }
```
aataraxiaa, line 104: Minor: We could avoid the creation of the event and return early by moving the conversion window check up:

```suggestion
        // Check if within conversion window
        guard isUserInConversionWindow(conversionWindowDays, enrollmentDate: experimentData.enrollmentDate) else { return }

        // Define event
        let event = event(for: subfeatureID, experimentData: experimentData, conversionWindowDays: conversionWindowDays, metric: metric, value: value)

        ExperimentConfig.fireFunction(event, .uniqueByNameAndParameters, false)
```

Sources/PixelExperimentKit/PixelExperimentKit.swift
```diff
@@ -92,13 +92,43 @@ extension PixelKit {
...
+    ///   - subfeatureID: Identifier for the subfeature associated with the experiment.
+    ///   - metric: The name of the metric being tracked (e.g., "searches").
+    ///   - conversionWindowDays: The range of days after enrollment during which the action is valid.
+    ///   - numberOfCalls: target number of actions required to fire the pixel.
+    ///
+    /// This function:
+    /// 1. Validates if the experiment is active.
+    /// 2. Ensures the user is within the specified conversion window.
+    /// 3. Tracks actions performed and sends the pixel once the target value is reached (if applicable).
+    public static func fireExperimentPixelWhenReachingNumberOfCalls(for subfeatureID: SubfeatureID,
```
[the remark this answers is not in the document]
aataraxiaa: I think we could include "threshold" in the name, e.g:

`fireExperimentPixelOnValueThreshold` or `fireExperimentPixelfThresholdReached`

Sources/PixelExperimentKit/PixelExperimentKit.swift
```diff
@@ -82,7 +82,7 @@ extension PixelKit {
     /// This function:
     /// 1. Validates if the experiment is active.
     /// 2. Ensures the user is within the specified conversion window.
-    /// 3. Tracks actions performed and sends the pixel once the target value is reached (if applicable).
+    /// 3. Sends the pixel if not sent before (unique by name and parameter)
```
aataraxiaa, line 85: Should we not be storing the pixel that was sent in this method so that we can ensure pixels are unique? (i.e not sent again if they were sent before). When the value being sent is unchanged, this method fires the same pixel on subsequent calls.

aataraxiaa (changes requested):
Thanks for the quick progress on this @SabrinaTardio. Left a few comments and tested this with my branch. The changes allow us to send pixels when the value param has changed. However, the API now also allows us to send the same pixel when the value is unchanged. I believe we still need to track the pixel that was sent so that we can ensure we don't send repeat pixels.

aataraxiaa (approved):
Thanks @SabrinaTardio, works great! 🥳

PR #1172, opened 2025-01-20
tds experiment metrics

SabrinaTardio:

**Required**:

Task/Issue URL: https://app.asana.com/0/1204186595873227/1208547600159201/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3839
macOS PR:  https://github.com/duckduckgo/macos-browser/pull/3755/files

**Optional**:

Tech Design URL: https://app.asana.com/0/1204186595873227/1209001242859416/f
CC:

**Description**: Adds pixel reporting for TDS override experiment

**Steps to test this PR**:
1. Look at platforms PRs

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


Sources/PixelExperimentKit/TDSOverrideExperimentMetrics.swift
```diff
@@ -0,0 +1,100 @@
...
+public enum TdsExperimentMetricType: String {
+    /// Metric triggered when the privacy toggle is used.
+    case privacyToggleUsed = "privacyToggleUsed"
+    /// Metric triggered after 2 quick refreshes.
+    case refresh2X = "2XRefresh"
+    /// Metric triggered after 3 quick refreshes.
+    case refresh3X = "3XRefresh"
+}
+
+public struct TDSOverrideExperimentMetrics {
```
jaceklyp, line 33: Im curious, is there any advantage of making all methods and vars static in this class? would we benefit anything from making it instance like? e.g. instantiating once specific code, passing featureFlagger in init, etc.?

jaceklyp (changes requested):

Sources/PixelExperimentKit/TDSOverrideExperimentMetrics.swift
SabrinaTardio: I was actually in two minds about this for a while.
In the end I opted for this approach because:
- Is consistent with the other PixelKit methods
- The featureFlagger is already passed to PixelExperimentKit at set up
- It makes it easier for it to be called from everywhere (for example internal packages that might not have access to the feature flagger)
- Testing the onDidDetectRefreshPattern that is an extension

jaceklyp: Sure, I understand. I don’t want to push for bigger changes at this stage, but personally, I see this as an anti-pattern. I think we’re collectively moving away from using singletons and static objects. While they’re convenient to access from anywhere, they make testing (mocking) much harder and tend to encourage this less maintainable style.

SabrinaTardio: Actually one of the reason I did it like this as I mentioned was for testability and I think that specifically for Pixel related things it makes sense to have static functions. PixelKit is new and has been implemented like that for a reason.

jaceklyp: Yeah, actually, PixelKit needs a refactor (it is somewhere on a roadmap), and one of the goals of this project is to transition it to an instance-based design. 😅

Sources/PageRefreshMonitor/PageRefreshMonitor.swift
```diff
@@ -70,6 +70,14 @@ public final class PageRefreshMonitor: PageRefreshMonitoring {
...
         resetIfURLChanged(to: url)
 
+        // Detect a refresh pattern if two refreshes occur within 12 seconds,
+        // triggering the experiment pixel (sent at most once per day).
+        // Detecting three refreshes within 21 seconds necessarily includes detecting two within 12 seconds,
+        // so clearing timestamps after detecting three refreshes has no impact on pixel sending since it is sent at most once per day.
+        if date.timeIntervalSince(refreshTimestamps.last ?? Date.distantPast) < 12.0 {
+            onDidDetectRefreshPattern(2)
+        }
+
```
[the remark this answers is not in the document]
SabrinaTardio: I think removing the refresh with flow the original logic (on 4 refresh within 20 seconds for example would send call the on onDidDetectRefreshPattern(3) twice which I think is not desirable).
Keeping it will not affect the pixel that can only be sent once a day anyway. (Most of the comment was to explain that). I wrote the simplest most efficient code to achieve the scope but your version is probably more readable so I ham happy to move to that (again without  removing the refresh)

jaceklyp (approved):
nice work :)

PR #1209, opened 2025-01-31
Passwords: Update save pixel with backfilled value

graeme:

**Required**:

Task/Issue URL: https://app.asana.com/0/1208273769335188/1209241071149720/f
iOS PR: https://github.com/duckduckgo/iOS/pull/3903
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3807

**Optional**:

Tech Design URL:
CC:

**Description**:

We're appending the backfilled parameter to the _update_ pixels, but not the _save_ pixels. So, for example, the case we're fixing for deviantart is not being captured if you're coming in fresh. Meaning, if you don't have credentials saved, we do backfill, but the pixel doesn't have backfilled=true, so we can't count it.

I think at some point we started discussing only about the _update_ pixel and that was misinterpreted as don't change anything for _save_, but that was not the intention.

The goal is to update the pixel firing logic both on macOS and iOS to match the above requirements

**Steps to test this PR**:

On both platforms...

1. Go to https://www.deviantart.com/users/login
2. Type in a username and hit Next
3. Type in a password and hit submit
4. You will see the Save password prompt with username and password
5. Check that you see `backfilled=true` in the save pixel on presentation, confirmation and dismissal
6. Save the password
7. Repeat steps 1 & 2
8. Type in a new password
9. You will see the Update password prompt with username and password
10. Check that you see `backfilled=true` in the update pixel on presentation, confirmation and dismissal

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [ ] macOS 10.15
* [ ] macOS 11
* [ ] macOS 12

---
###### Internal references:


amddg44 (approved):
LGTM!

PR #1213, opened 2025-02-05
VPN is sometimes stopped twice

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/1207603085593419/1209322889682604/f

iOS PR: https://github.com/duckduckgo/iOS/pull/3928
macOS PR: https://github.com/duckduckgo/macos-browser/pull/3829

## Description

Fixes an issue where the tunnel is being stopped twice.

## Testing

Please refer to the platform-specific PRs for testing instructions.

---
###### Internal references:

### DIFF
--- a/Sources/NetworkProtection/PacketTunnelProvider.swift
+++ b/Sources/NetworkProtection/PacketTunnelProvider.swift
@@ -700,17 +700,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             }
 
             Logger.networkProtection.error("🔴 Stopping VPN due to no auth token")
-
-            // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
-            if let wrappedError = wrapped(error: error) {
-                // Wait for the provider to complete its pixel request.
-                try? await Task.sleep(interval: .seconds(2))
-                throw wrappedError
-            } else {
-                // Wait for the provider to complete its pixel request.
-                try? await Task.sleep(interval: .seconds(2))
-                throw error
-            }
+            throw error
         }
 
         do {
@@ -739,17 +729,7 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             self.knownFailureStore.lastKnownFailure = KnownFailure(error)
 
             providerEvents.fire(.tunnelStartAttempt(.failure(error)))
-
-            // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
-            if let wrappedError = wrapped(error: error) {
-                // Wait for the provider to complete its pixel request.
-                try? await Task.sleep(interval: .seconds(2))
-                throw wrappedError
-            } else {
-                // Wait for the provider to complete its pixel request.
-                try? await Task.sleep(interval: .seconds(2))
-                throw error
-            }
+            throw error
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -700,7 +700,6 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             }
 
             Logger.networkProtection.error("🔴 Stopping VPN due to no auth token")
-            await cancelTunnel(with: TunnelError.startingTunnelWithoutAuthToken)
 
             // Check that the error is valid and able to be re-thrown to the OS before shutting the tunnel down
             if let wrappedError = wrapped(error: error) {
```
diegoreymendez, line 705: Should we also remove this logic and just throw the error?  I'm feeling like it may be a good idea to clean this up and see if the above fix resolves both issues.

I don't think this has resolve the issue - there are still a good number of reports on Sentry.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -700,7 +700,6 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
             }
 
             Logger.networkProtection.error("🔴 Stopping VPN due to no auth token")
-            await cancelTunnel(with: TunnelError.startingTunnelWithoutAuthToken)
```
diegoreymendez, line 703: No need to cancel the tunnel if we're throwing an error below, which is the right way to cancel startup.

samsymons (approved):

Sources/NetworkProtection/PacketTunnelProvider.swift
diegoreymendez: Removed the outdated code.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -111,14 +112,16 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
-                return "Missing auth token at startup"
+            case .startingTunnelWithoutAuthToken(let internalError):
+                return "Missing auth token at startup: \(internalError)"
             case .vpnAccessRevoked:
                 return "VPN disconnected due to expired subscription"
             case .couldNotGenerateTunnelConfiguration(let internalError):
                 return "Failed to generate a tunnel configuration: \(internalError.localizedDescription)"
             case .simulateTunnelFailureError:
                 return "Simulated a tunnel error as requested"
+            case .tokenReset:
```
diegoreymendez, line 123: Differentiating a specific case that was previously reported as `startingTunnelWithoutAuthToken`.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -581,13 +586,13 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
...
             } catch {
-                throw TunnelError.startingTunnelWithoutAuthToken
+                throw TunnelError.startingTunnelWithoutAuthToken(internalError: error)
             }
         case .reset:
             // This case should in theory not be possible, but it's ideal to have this in place
             // in case an error in the controller on the client side allows it.
-            try await tokenHandler.removeToken()
-            throw TunnelError.startingTunnelWithoutAuthToken
+            try? await tokenHandler.removeToken()
```
diegoreymendez, line 594: Even if this fails, this isn't the real failure reason - the real failure reason is the token is being removed which, as described above, is an abnormal scenario.  The removal failing shouldn't be the error we get.

Sources/NetworkProtection/PacketTunnelProvider.swift
```diff
@@ -603,22 +608,22 @@ open class PacketTunnelProvider: NEPacketTunnelProvider {
                 try await tokenHandler.refreshToken()
             } catch {
                 Logger.networkProtection.fault("Error force-refreshing token container: \(error, privacy: .public)\n \(newTokenContainer.refreshToken, privacy: .public)")
-                throw TunnelError.startingTunnelWithoutAuthToken
+                throw TunnelError.startingTunnelWithoutAuthToken(internalError: error)
```
diegoreymendez, line 611: Just taking advantage that there's underlying error info.

Sources/NetworkProtection/PacketTunnelProvider.swift
samsymons: Nice, yeah good catch.

samsymons (approved):

PR #1216, opened 2025-02-05
Dead token recovery

federicocappelli:
Task/Issue URL: https://app.asana.com/0/1205842942115003/1209252725803000
iOS PR: no
macOS PR: no
What kind of version bump will this require?: none

**Description**:

This PR introduces improvements in the "dead token" detection and recovery for Auth V2.
A deat token happens when the refresh has expired.
The only thing we can do is log out the user and try to recover the subscription, if it was purchased via the Apple store, or show an error to the user for Stripe.

**Steps to test this PR**:
The code is not used yet, check the unit tests

---
###### Internal references:

### DIFF
--- a/Sources/Networking/Auth/OAuthClient.swift
+++ b/Sources/Networking/Auth/OAuthClient.swift
@@ -129,6 +129,13 @@ final public class DefaultOAuthClient: OAuthClient {
         static let clientID = "f4311287-0121-40e6-8bbd-85c36daf1837"
         static let redirectURI = "com.duckduckgo:/authcb"
         static let availableScopes = [ "privacypro" ]
+
+#if DEBUG
+        /// The seconds before the expiry date when we consider a token effectively expired
+        static let tokenExpiryBufferInterval: Int = 30
+#else
+        static let tokenExpiryBufferInterval: Int = 60*10
+#endif
     }
 
     // MARK: -
@@ -232,13 +239,8 @@ final public class DefaultOAuthClient: OAuthClient {
             let tokenExpiryDate = localTokenContainer.decodedAccessToken.exp.value
             Logger.OAuthClient.debug("Local tokens found, expiry: \(tokenExpiryDate, privacy: .public)")
 
-#if DEBUG
-            let expiresSoon = false
-#else
-            // Expires in less than 10 minutes, check only in release, the staging token expires every 4 minutes
-            let expiresSoon = tokenExpiryDate.minutesSinceNow() < 10
-#endif
-
+            // If the token expires in less than `Constants.tokenExpiryBufferInterval` minutes we treat it as already expired
+            let expiresSoon = tokenExpiryDate.secondsFromNow() < Constants.tokenExpiryBufferInterval
             if localTokenContainer.decodedAccessToken.isExpired() || expiresSoon {
                 Logger.OAuthClient.debug("Local access token is expired, refreshing it")
                 return try await getTokens(policy: .localForceRefresh)
--- a/Sources/Networking/Auth/OAuthClient.swift
+++ b/Sources/Networking/Auth/OAuthClient.swift
@@ -26,6 +26,7 @@ public enum OAuthClientError: Error, LocalizedError, Equatable {
     case unauthenticated
     /// When both access token and refresh token are expired
     case refreshTokenExpired
+    case invalidTokenRequest
 
     public var errorDescription: String? {
         switch self {
@@ -39,6 +40,8 @@ public enum OAuthClientError: Error, LocalizedError, Equatable {
             return "The account is not authenticated, please re-authenticate"
         case .refreshTokenExpired:
             return "The refresh token is expired, the token is unrecoverable please re-authenticate"
+        case .invalidTokenRequest:
+            return "Invalid token request"
         }
     }
 }
@@ -204,7 +207,15 @@ final public class DefaultOAuthClient: OAuthClient {
     public func getTokens(policy: AuthTokensCachePolicy) async throws -> TokenContainer {
         let localTokenContainer = tokenStorage.tokenContainer
 
+        if policy != .local,
+           let localTokenContainer,
+            localTokenContainer.decodedRefreshToken.isExpired() {
+            // The refresh token is expired, the token is un-refreshable
+            throw OAuthClientError.refreshTokenExpired
+        }
+
         switch policy {
+
         case .local:
             guard let localTokenContainer else {
                 Logger.OAuthClient.debug("Tokens not found")
@@ -212,18 +223,29 @@ final public class DefaultOAuthClient: OAuthClient {
             }
             Logger.OAuthClient.debug("Local tokens found, expiry: \(localTokenContainer.decodedAccessToken.exp.value, privacy: .public)")
             return localTokenContainer
+
         case .localValid:
             guard let localTokenContainer else {
                 Logger.OAuthClient.debug("Tokens not found")
                 throw OAuthClientError.missingTokens
             }
-            Logger.OAuthClient.debug("Local tokens found, expiry: \(localTokenContainer.decodedAccessToken.exp.value, privacy: .public)")
-            if localTokenContainer.decodedAccessToken.isExpired() {
+            let tokenExpiryDate = localTokenContainer.decodedAccessToken.exp.value
+            Logger.OAuthClient.debug("Local tokens found, expiry: \(tokenExpiryDate, privacy: .public)")
+
+#if DEBUG
+            let expiresSoon = false
+#else
+            // Expires in less than 10 minutes, check only in release, the staging token expires every 4 minutes
+            let expiresSoon = tokenExpiryDate.minutesSinceNow() < 10
+#endif
+
+            if localTokenContainer.decodedAccessToken.isExpired() || expiresSoon {
                 Logger.OAuthClient.debug("Local access token is expired, refreshing it")
                 return try await getTokens(policy: .localForceRefresh)
             } else {
                 return localTokenContainer
             }
+
         case .localForceRefresh:
             guard let refreshToken = localTokenContainer?.refreshToken else {
                 Logger.OAuthClient.debug("Refresh token not found")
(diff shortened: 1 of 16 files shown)


Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -204,26 +207,45 @@ final public class DefaultOAuthClient: OAuthClient {
     public func getTokens(policy: AuthTokensCachePolicy) async throws -> TokenContainer {
         let localTokenContainer = tokenStorage.tokenContainer
 
+        if policy != .local,
```
federicocappelli, line 210: we now check the refresh token validity every time we return a token

Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -204,26 +207,45 @@ final public class DefaultOAuthClient: OAuthClient {
...
             }
-            Logger.OAuthClient.debug("Local tokens found, expiry: \(localTokenContainer.decodedAccessToken.exp.value, privacy: .public)")
-            if localTokenContainer.decodedAccessToken.isExpired() {
+            let tokenExpiryDate = localTokenContainer.decodedAccessToken.exp.value
+            Logger.OAuthClient.debug("Local tokens found, expiry: \(tokenExpiryDate, privacy: .public)")
+
+#if DEBUG
+            let expiresSoon = false
+#else
+            // Expires in less than 10 minutes, check only in release, the staging token expires every 4 minutes
```
federicocappelli, line 238: I've added a local check for the token expiry date, if it's less than 10 minutes we treat it as expired

Sources/Networking/Auth/OAuthClient.swift
```diff
@@ -237,11 +259,12 @@ final public class DefaultOAuthClient: OAuthClient {
                 return refreshedTokens
             } catch OAuthServiceError.authAPIError(let code) where code == OAuthRequest.BodyErrorCode.invalidTokenRequest {
                 Logger.OAuthClient.error("Failed to refresh token: invalidTokenRequest")
-                throw OAuthClientError.refreshTokenExpired
+                throw OAuthClientError.invalidTokenRequest
```
federicocappelli, line 262: here I maintain a similar error to the one sent by the API. `.invalidTokenRequest` can happen for other reasons than a dead token, for example when we try to use a staging token in production.

Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+
+import Foundation
+import Networking
+import os.log
+
+@available(macOS 12.0, *)
+public struct DeadTokenRecoverer {
+
+    private static var recoveryAttempted: Int = 0
```
federicocappelli, line 26: In case the BE has a problem and always returns a `OAuthClientError.invalidTokenRequest` every time we request a token, we don't want to enter an infinite loop, so we must to be sure that we are not already in a recovery state.
This state helps with that and it's not a bool because multiple, parallel API requests could use this utility and we want to be sure we don't have concurrency issues.

Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -0,0 +1,49 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import Networking
+import os.log
+
+@available(macOS 12.0, *)
+public struct DeadTokenRecoverer {
```
federicocappelli, line 24: This utility must be outside the subscription manager because requires Apple recovery flow, but we don't want to re-implement it every time, so the dead token recovery business logic is all here and can be used by the main apps.
This utility main function is used in the SubscriptionManagerV2 autoRecoveryHandler block

Tests/SubscriptionTests/Managers/SubscriptionManagerV2Tests.swift
```diff
@@ -28,20 +28,29 @@ class SubscriptionManagerV2Tests: XCTestCase {
...
+        mockAppStoreRestoreFlowV2 = AppStoreRestoreFlowMockV2()
 
         subscriptionManager = DefaultSubscriptionManagerV2(
             storePurchaseManager: mockStorePurchaseManager,
             oAuthClient: mockOAuthClient,
             subscriptionEndpointService: mockSubscriptionEndpointService,
             subscriptionEnvironment: SubscriptionEnvironment(serviceEnvironment: .staging, purchasePlatform: .stripe),
-            pixelHandler: { _ in }
+            pixelHandler: { _ in },
+            autoRecoveryHandler: {
```
federicocappelli, line 48: this is an example of the new handler implementation, ignore the `overrideTokenResponse ` that is used only for testing purposes

Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -305,33 +327,27 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
...
+            do { return try await attemptTokenRecovery() } catch { throw error }
         } catch {
             throw SubscriptionManagerError.tokenUnavailable(error: error)
         }
     }
 
-    /// If the client succeeds in making a refresh request but does not get the response, then the second refresh request will fail with `invalidTokenRequest` and the stored token will become unusable and un-refreshable.
-    private func throwAppropriateDeadTokenError() async throws -> TokenContainer {
-        Logger.subscription.fault("Dead token detected")
+    func attemptTokenRecovery() async throws -> TokenContainer {
```
federicocappelli, line 343: this is where the autoRecoveryHandler is used

Sources/Subscription/DeadTokenRecoverer.swift
miasma13, line 26: I think `recoveryAttempted` would more fit a boolean state, wdyt about `recoveryAttemptCount`?

Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -21,22 +21,37 @@ import Common
...
+    public var errorDescription: String? {
+        switch self {
+        case .tokenUnavailable(error: let error):
+            "Token unavailable: \(String(describing: error))"
+        case .confirmationHasInvalidSubscription:
+            "Confirmation has an invalid subscription"
+        case .noProductsFound:
+            "No products found"
+        case .tokenUnRefreshable:
+            "Token is not refreshable, the account will be log out"
```
miasma13, line 52: "...logged out"

Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -103,7 +121,9 @@ public protocol SubscriptionManagerV2: SubscriptionTokenProvider {
     /// Get a token container accordingly to the policy
     /// - Parameter policy: The policy that will be used to get the token, it effects the tokens source and validity
     /// - Returns: The TokenContainer
-    /// - Throws: OAuthClientError.deadToken if the token is unrecoverable. SubscriptionEndpointServiceError.noData if the token is not available.
+    /// - Throws: A `SubscriptionManagerError`.
+    ///     `tokenUnRefreshable` if the token is cannot be refreshed, typically due to an expired refresh token.
```
miasma13, line 125: "if the token cannot be..."

Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -291,10 +314,9 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
     // MARK: -
 
     @discardableResult public func getTokenContainer(policy: AuthTokensCachePolicy) async throws -> TokenContainer {
+        Logger.subscription.debug("Get tokens \(policy.description, privacy: .public)")
         do {
-            Logger.subscription.debug("Get tokens \(policy.description, privacy: .public)")
-
-            let referenceCachedTokenContainer = try? await oAuthClient.getTokens(policy: .local) // the currently stored one
+            let referenceCachedTokenContainer = try? await oAuthClient.getTokens(policy: .local)
```
miasma13, line 319: What do you think about replacing `reference` with `current` in both  `referenceCachedTokenContainer` and `referenceCachedEntitlements`?

Sources/Subscription/Managers/SubscriptionManagerV2.swift
```diff
@@ -305,33 +327,27 @@ public final class DefaultSubscriptionManagerV2: SubscriptionManagerV2 {
                 NotificationCenter.default.post(name: .entitlementsDidChange, object: self, userInfo: [UserDefaultsCacheKey.subscriptionEntitlements: newEntitlements])
             }
 
-            if referenceCachedTokenContainer == nil { // new login
+            // Send "accountDidSignIn" notification when login changes
```
miasma13, line 330: It is hard to tell without testing it but seems that the order of sending notification should be reversed, in scenario when signing in, the `entitlementsDidChange` notification before `accountDidSignIn` one does not make much sense.

Sources/Subscription/DeadTokenRecoverer.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    private static var recoveryAttempted: Int = 0
+
+    public static func recoverDeadToken(endpointService: any SubscriptionEndpointServiceV2,
```
miasma13, line 28: Two things:
1. I am thinking about more descriptive name for it as it does very specific thing. I am missing that this is an attempt (and may fail) and how the recovery is made. Suggestions: 
- `attemptDeadTokenRecoveryFromPastPurchase(endpointService: any SubscriptionEndpointServiceV2, restoreFlow: any AppStoreRestoreFlowV2)`
- `attemptDeadTokenRecovery(using restoreFlow: any AppStoreRestoreFlowV2, endpointService: any SubscriptionEndpointServiceV2)`
2. To be wary for the "real" implementation if the function can potentially be invoked multiple times in parallel without synchronisation.

miasma13 (approved):
Overall LGTM, just couple of comments

Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: Correct, it was a boolean before :D Fixed

Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: fixed

Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: fixed

Sources/Subscription/Managers/SubscriptionManagerV2.swift
federicocappelli: fixed

Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: fixed

Sources/Subscription/DeadTokenRecoverer.swift
federicocappelli: 1. What about `DeadTokenRecoverer.attemptRecoveryFromPastPurchase(...)`?
2. In theory, `recoveryAttemptCount` in the struct is the safeguard for that. Only one recovery operation at the time will be executed, the other ones will throw a `.tokenUnRefreshable`

Sources/Networking/Auth/OAuthClient.swift
federicocappelli: changed!

